Issue #175, opened 2017-10-06
Stop camera while not needed


Gallore:
Stop taking pictures while it's not needed even if the lid is open. 

Cases are:
- during conversion 
- if user signed out / no user is signed in
- if device is not homed

PR #189, opened 2017-10-23
F unnecessary pics

clemniem:
Tested on my machine.


make-ing:
Seems fine to me.

About your TODO: I think this is just legacy code and could be implemented with all diffrent functions as you suggested.

clemniem:
@make-ing ok, thanks for the review! 
I think @Gallore will merge. I don't mess with the develop branch while abroad :)

clemniem:
@Gallore All states from #175 are now implemented and tested on my machine.

PR #201, opened 2017-11-03
f_analytics

clemniem:
Merge f_analytics into develop

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -80,6 +80,9 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 		self.print_progress_last = -1
 		self.slicing_progress_last = -1
 		self._logger = mrb_logger("octoprint.plugins.mrbeam")
+		self._hostname = self.getHostname()
+		self._mbSerialnumber = self.getMrBeamSerial()
+
 
 	# inside initialize() OctoPrint is already loaded, not assured during __init__()!
 	def initialize(self):
@@ -87,8 +90,6 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 		self.laserCutterProfileManager = laserCutterProfileManager()
 		self._logger = mrb_logger("octoprint.plugins.mrbeam")
 		self._branch = self.getBranch()
-		self._hostname = self.getHostname()
-		self._mbSerialnumber = self.getMrBeamSerial()
 		self._octopi_info = self.get_octopi_info()
 		self._serial = self.getMrBeamSerial()
 		self._do_initial_log()
@@ -1339,13 +1340,15 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 		ALSO: Checks if hostname not set and initializes otherwise.
 		:return: String hostname
 		"""
-		if not self._hostname:
+		try:
+			return self._hostname
+		except AttributeError:
 			try:
 				self._hostname = socket.gethostname()
 			except:
 				self._logger.exception("Exception while reading hostname.")
 				pass
-		return self._hostname
+			return self._hostname
 
 	def getDisplayName(self, hostName):
 		code = None
@@ -1358,7 +1361,9 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 			return name.format(hostName)
 
 	def getMrBeamSerial(self):
-		if not self._mbSerialnumber:
+		try:
+			return self._mbSerialnumber
+		except AttributeError:
 			self._mbSerialnumber = "{pi_serial}-{device_series}".format(
 			pi_serial=self.getPiSerial(),
 			device_series=self._get_val_from_device_info('device_series'))
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -80,9 +80,10 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 		self.print_progress_last = -1
 		self.slicing_progress_last = -1
 		self._logger = mrb_logger("octoprint.plugins.mrbeam")
-		self._hostname = self.getHostname()
-		self._mbSerialnumber = self.getMrBeamSerial()
-
+		self._hostname = None
+		self._mbSerialnumber = None
+		self._setHostname()
+		self._setMbSerialnumber()
 
 	# inside initialize() OctoPrint is already loaded, not assured during __init__()!
 	def initialize(self):
@@ -1341,21 +1342,22 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 		else:
 			return None, None
 
+	def _setHostname(self):
+		try:
+			self._hostname = socket.gethostname()
+		except:
+			self._logger.exception("Exception while reading hostname.")
+			pass
+
 	def getHostname(self):
 		"""
 		Returns device hostname like 'MrBeam2-F930'.
-		ALSO: Checks if hostname not set and initializes otherwise.
 		:return: String hostname
 		"""
-		try:
-			return self._hostname
-		except AttributeError:
-			try:
-				self._hostname = socket.gethostname()
-			except:
-				self._logger.exception("Exception while reading hostname.")
-				pass
-			return self._hostname
+		if self._hostname is None:
+			self._logger.warning('Hostname was not initialized. Doing it now.')
+			self._setHostname()
+		return self._hostname
 
 	def getDisplayName(self, hostName):
 		code = None
@@ -1367,13 +1369,15 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 		else:
 			return name.format(hostName)
 
-	def getMrBeamSerial(self):
-		try:
-			return self._mbSerialnumber
-		except AttributeError:
-			self._mbSerialnumber = "{pi_serial}-{device_series}".format(
+	def _setMbSerialnumber(self):
+		self._mbSerialnumber = "{pi_serial}-{device_series}".format(
 			pi_serial=self.getPiSerial(),
 			device_series=self._get_val_from_device_info('device_series'))
+
+	def getMrBeamSerial(self):
+		if self._mbSerialnumber is None:
+			self._logger.warning('Serial Number was not initialized. Doing it now.')
+			self._setMbSerialnumber()
 		return self._mbSerialnumber
 
 	def getPiSerial(self):
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -1106,6 +1106,8 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 		pic_settings['calibMarkers'] = newMarkers
 		pic_settings['calibration_updated'] = True
 
+		self._analytics_handler.write_cam_update(newMarkers,newCorners)
+
 		self._logger.debug('picSettings new to save: {}'.format(pic_settings))
 		self._save_profile(pic_settings_path,pic_settings)
 
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -166,14 +166,19 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 				# ),
 				iobeam_disable_warnings = False
 			),
+			# TODO rename analyticsEnabled and put in analytics-dict
 			analyticsEnabled=False,  # frontend analytics Mixpanel
 			analyticsfolder="analytics",  # laser job analytics base folder (.octoprint/...)
+			analytics=dict(
+				job_analytics = True,
+				cam_analytics = True,
+				folder = 'analytics'
+			),
 			cam=dict(
 				enabled=True,
 				image_correction_enabled = True,
 				cam_img_width = image_default_width,
 				cam_img_height = image_default_height,
-				# todo CLEM add NPZ folder and pic_settings, calib output folder etc.
 				frontendUrl="/downloads/files/local/cam/beam-cam.jpg",
 				localFilePath="cam/beam-cam.jpg",
 				localUndistImage="cam/undistorted.jpg",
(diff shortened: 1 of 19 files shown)


octoprint_mrbeam/__init__.py
```diff
@@ -1335,16 +1341,20 @@ def _getCurrentFile(self):
...
+		Returns device hostname like 'MrBeam2-F930'.
+		ALSO: Checks if hostname not set and initializes otherwise.
+		:return: String hostname
+		"""
 		try:
-			hostname = socket.gethostname()
-		except:
-			hostname = ERROR.HOSTNAME
-		return hostname
+			return self._hostname
```
andywerner: I don't get this. What would trigger an AttributeError given that you initialize self._hostname in __init__() (line 83) Therefore: How would _hostname ever hold a value?

octoprint_mrbeam/__init__.py
```diff
@@ -1357,16 +1367,20 @@ def getDisplayName(self, hostName):
 			return name.format(hostName)
 
 	def getMrBeamSerial(self):
-		return "{pi_serial}-{device_series}".format(
+		try:
+			return self._mbSerialnumber
```
andywerner: Same here.

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -21,14 +32,45 @@ def __init__(self, event_bus, settings):
...
+		self._deviceinfo_log_version = 2
+		self._dust_log_version = 2
+		self._cam_event_log_version = 2
+
+
+		analyticsfolder = os.path.join(self._settings.getBaseFolder("base"), self._settings.get(['analytics','folder']))
 		if not os.path.isdir(analyticsfolder):
 			os.makedirs(analyticsfolder)
 
 		self._jsonfile = os.path.join(analyticsfolder, "analytics_log.json")
```
andywerner: Can we pls have this ("analytics_log.json") in a CONSTANT or even better from settings/config.yaml. 
(I know, it's not your code, Clemens...)

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -21,14 +32,45 @@ def __init__(self, event_bus, settings):
...
-		self._initjsonfile()
+		if not os.path.isfile(self._jsonfile):
+			self._init_jsonfile()
+
+		if self._analyticsOn:
+			# TODO ANDY : Check if this is how you want it or maybe print it out at every startup
+			# check if <two days> have passed and software should be written away
+			self._days_passed_since_last_log = self._days_passed(self._jsonfile)
+			self._logger.debug('Days since last edit: {}'.format(self._days_passed_since_last_log))
+			if self._days_passed_since_last_log > 2:
```
andywerner: Pls have this 2 (days) as a CONSTANT here.

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -21,14 +32,45 @@ def __init__(self, event_bus, settings):
...
 
 		self._jsonfile = os.path.join(analyticsfolder, "analytics_log.json")
-		self._initjsonfile()
+		if not os.path.isfile(self._jsonfile):
+			self._init_jsonfile()
+
+		if self._analyticsOn:
+			# TODO ANDY : Check if this is how you want it or maybe print it out at every startup
+			# check if <two days> have passed and software should be written away
+			self._days_passed_since_last_log = self._days_passed(self._jsonfile)
```
andywerner: Why do you need _days_passed_since_last_log as a member (self.)
I suggest to either change it to a local variable (without self) or initialize it a little up there and set it to -1

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -40,80 +82,339 @@ def _subscribe(self):
...
-					'hostname': self._gethostname(),
-					'timestamp': time.time()
-				}
-				json.dump(data, f)
-				f.write('\n')
 
 	@staticmethod
-	def _getserialnumber():
+	def _getLaserHeadVersion():
+		# TODO ANDY tell CLEM how to get Real laser_head_id :)
```
andywerner: Not yet supported by iobeam. Let's use a value that is obviously not a real version like 99, None, -1 or just skip this information for now.

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -40,80 +82,339 @@ def _subscribe(self):
...
+		lm_date = datetime.utcfromtimestamp(lm_ts)
+
+		now_date = datetime.utcnow()
+		days_passed = (now_date-lm_date).days
+
+		return days_passed
+
+	def _write_current_software_status(self):
+		try:
+			# TODO ANDY get all software statuses
```
andywerner: yepp. let's do this later.

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -40,80 +82,339 @@ def _subscribe(self):
...
+		self._write_jobevent(ak.PRINT_STARTED, {ak.FILENAME: filename})
+
+	def _init_collectors(self):
+		self._current_dust_collector = ValueCollector('DustColl')
+		self._current_intensity_collector = ValueCollector('IntensityColl')
+		self._current_lasertemp_collector = ValueCollector('TempColl')
 
 	def _event_print_paused(self, event, payload):
-		self.write_event('jobevent', 'print_paused', 1)
+		# TODO CLEM add how it has been paused (lid_opened during job, frontend, onebutton)
```
andywerner: Nicht ganz sicher, ob meine Annahme hier stimmt... 

Cooling: payload holds some information if it was a cooling_pause or not. 
Lid/Button: Currently there is no way to know other than checking the current state: _mrbeam_plugin_implementation._ioBeam .is_interlock_closed()

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -40,80 +82,339 @@ def _subscribe(self):
...
+			self._write_jobevent(ak.PRINT_DONE)
+			self._write_collectors()
+
+	def _write_collectors(self):
+		self._write_jobevent(ak.DUST_SUM,payload=self._current_dust_collector.getSummary())
+		self._write_jobevent(ak.INTENSITY_SUM,payload=self._current_intensity_collector.getSummary())
+		self._write_jobevent(ak.LASERTEMP_SUM, payload=self._current_lasertemp_collector.getSummary())
+
+	def _cleanup(self,successfull):
+		# TODO check if resetting job_id etc to None makes sense
```
andywerner: I guess it does?

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -40,80 +82,339 @@ def _subscribe(self):
 		self._event_bus.subscribe(MrBeamEvents.PRINT_PROGRESS, self._event_print_progress)
 		self._event_bus.subscribe(MrBeamEvents.LASER_COOLING_PAUSE, self._event_laser_cooling_pause)
```
andywerner, line 82: Wie ist das, schmeissen wir nen LASER_COOLING_PAUSE und regulären PAUSE event? Bin mir grad nicht so sicher... Wenn du das schon gecheckt oder getestet hast, dann passts.

octoprint_mrbeam/analytics/analytics_keys.py
```diff
@@ -0,0 +1,72 @@
+class AnalyticsKeys(object):
```
andywerner, line 1: nice!!!! I like this

octoprint_mrbeam/analytics/value_collector.py
```diff
@@ -0,0 +1,50 @@
...
+				'max': max(self.valueList),
+				'25p': np.percentile(arr,25),
+				'75p': np.percentile(arr,75),
+				'std': np.std(arr),
+				'count': len(self.valueList)
+			}
+		else:
+			descDict = {'count':count}
+
+		# make all values float for json.dump()-compability
```
andywerner, line 39: Would it make a difference, if we'd used something like JSON.stringify() ? Probably not...

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -40,80 +82,339 @@ def _subscribe(self):
...
 
 	def _append_data_to_file(self, data):
-		with open(self._jsonfile, 'a') as f:
-			json.dump(data, f)
-			f.write('\n')
+		try:
+			if not os.path.isfile(self._jsonfile):
+				self._init_jsonfile()
+			with open(self._jsonfile, 'a') as f:
+				json.dump(data, f)
```
andywerner: Could we pls get the json as string, append the newline and then write it in one pseudo-atomic operation. Appears to be more fail safe than this.
(I know, not your code, Clemens)

octoprint_mrbeam/gcodegenerator/converter.py
```diff
@@ -60,14 +60,16 @@ def __init__(self, params, model_path):
...
+		_mrbeam_plugin_implementation._analytics_handler.store_conversion_details(self.options)
 
 	def setoptions(self, opts):
 		# set default values if option is missing
 		# self._log.info("opts: %s" % opts)
 		for key in self.options.keys():
 			if key in opts:
 				self.options[key] = opts[key]
-				if(key == "vector"):
+				if key == "vector":
```
andywerner, line 72: :-)

octoprint_mrbeam/iobeam/dust_manager.py
```diff
@@ -203,15 +206,39 @@ def _send_fan_command(self, command, wait_time=-1.0, max_retries=-1):
...
-		_mrbeam_plugin_implementation._analytics_handler.add_dust_log(data)
+	def _send_dust_to_analytics(self,val):
+		"""
+		Sends dust value periodically to analytics_handler to get overall stats and dust profile.
+		:param val: measured dust value
+		:return:
+		"""
+		_mrbeam_plugin_implementation._analytics_handler.add_dust_value(val)
+
+	def _send_final_dust_to_analytics(self, dust_start, dust_start_ts, dust_end, dust_end_ts):
```
andywerner: This method should not be here but in the analytics package. (Eg. analytics_handler). 
All this method does is data preparation/conversion for analytics. Therefore it should be in the analytics package. Also AnalyticsKeys (as ak) should only be used WITHIN the analytics package IMO.
(I know, the capsulation in/around the iobeam package here is not really textbook style...)

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,7 +37,7 @@ def get_update_information(self):
 		set_info_rpiws281x(self, tier)
 #		set_info_testplugin(self, tier) # See function definition for more details
 
-	# _logger(self).debug("MrBeam Plugin provides this config (might be overridden by settings!):\n%s", yaml.dump(sw_update_config, width=50000).strip())
+	_logger(self).debug("MrBeam Plugin provides this config (might be overridden by settings!):\n%s", yaml.dump(sw_update_config, width=50000).strip())
```
andywerner: Please uncomment this line again. (So that it is a comment and doesn't get executed) It's output is useful for debugging but too misleading to keep it in there.
It's misleading because anything in confi.yamel overrides these values but you wouldn't see it in this log output.

andywerner (changes requested):
I have some change requests. See detailed comments.

octoprint_mrbeam/__init__.py
clemniem: I don't know either.

clemniem: Before, there was the if else, like this:
```python
if not self._hostname:
    self._hostname = socket.gethostname()

return self._hostname
```

But this gave me an attribute error, because you cannot check with 'if' for an unset variable.

octoprint_mrbeam/analytics/analytics_handler.py
clemniem: done

octoprint_mrbeam/analytics/analytics_handler.py
clemniem: done

octoprint_mrbeam/analytics/analytics_handler.py
clemniem: done

octoprint_mrbeam/analytics/analytics_handler.py
clemniem: es wird bei einer Pause immer folgende events gesendet:
PAUSE_START
COOLING_START
COOLING_END
PAUSE_END

(Das sind nicht die richtigen Namen aber die Funktion ist soweit erkennbar denke ich.)

octoprint_mrbeam/analytics/analytics_handler.py
clemniem: None it is.

octoprint_mrbeam/analytics/analytics_handler.py
clemniem: Ja, ich glaube das ist auch erstmal nicht ganz soo wichtig, außer du sagst du willst diese Information haben.

octoprint_mrbeam/analytics/analytics_handler.py
clemniem: yep

octoprint_mrbeam/analytics/value_collector.py
clemniem: hm, but then we still need to round it, because otherwise we would have around 10+ digits after the comma.

octoprint_mrbeam/iobeam/dust_manager.py
clemniem: done

octoprint_mrbeam/software_update_information.py
clemniem: done

octoprint_mrbeam/gcodegenerator/converter.py
clemniem: :)

andywerner (approved):

PR #210, opened 2017-11-08
F key move design

clemniem:
implemented key moving design instead of laserhead, cleaned up code a little bit.


hungerpirat:
awesome.

andywerner (approved):

Issue #159, opened 2017-10-02
arrow keys error after limit hit


clemniem:
If the arrow keys are pressed, the mr beam is moved.
if you move him to one end of the working area there is an unhandled error message due to the limit hit. For example:
- home mr beam
- press 3 times to the right 
--> lots of error messages.
They shouldnt pop up again and again. And the arrows probably shouldnt move the mr beam. 
Maybe shift+arrow: moves mr beam
arrow: moves design

clemniem:
cant do homing until all error messages are clicked away! that is really annoying as well.

Gallore:
We need to catch the limit hit before it happens.

hungerpirat:
already fixed 2017.

PR #212, opened 2017-11-09
B 159 limit hit

clemniem:
No more limit hit errors. 
Implemented sendSaveJogCommand to check for limit hit before sending jogCommand.


clemniem:
#159 fixed

andywerner (changes requested):
I assume this has been replaced by something much better.
With this CR the arrow keys still move the laser head. But in develop we already have a version that moves the design (==preferred.)
Please let me know if I'm wrong here.

PR #323, opened 2018-08-29
B 28 graceful error handling

hungerpirat:
try catch on loading errors somehow does not work on this asynchronous callback stuff. Additionally Snapsvg does not feel bad in any way while parsing binary content as if it was xml. 
No exceptions, no errors - result is just one big text node with crap inside.
My error detection checks for non-printable characters which are usually only present in binary files.


andywerner:
manually cherry-picked

PR #332, opened 2018-09-17
Hotfix check disk space during conversion

andywerner:
Checks free disk space during conversion and if value is below min_required_disk_space (from config.yaml) the conversion will end with an OutOfSpaceException which triggers a nice explanation on frontend side.


andywerner:
Please Squash before merge!!!!

PR #388, opened 2019-02-08
Save information about design files and disk space (analytics)

irlaec:


octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -259,6 +263,24 @@ def _event_ip_addresses(self):
 		except:
 			self._logger.exception('Exception when recording the IP addresses')
 
+	def _event_disk_space(self):
+		try:
+			statvfs = os.statvfs('/')
+			total_space = statvfs.f_frsize * statvfs.f_blocks
+			available_space = statvfs.f_frsize * statvfs.f_bavail  # Available space for non-super users
+			free_space = statvfs.f_frsize * statvfs.f_bfree  # Free space for super users
```
andywerner: This seems unused. However, let's add this to the disk_space payload object so that we have it. Actually this is the most interesting number.

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -383,7 +405,7 @@ def store_conversion_details(self, details):
 					data.update(details['material'])
 					self._store_conversion_details(eventname,payload=data)
 
-				if 'vector' in details and details['vector'] != []:
+				if 'vector' in details and details['vector']:
```
andywerner, line 407: :-) yes!

andywerner (changes requested):

octoprint_mrbeam/analytics/analytics_handler.py
irlaec: The free_space was the available_space (for the user) + the available space for the super user. It was not used so now it's removed

andywerner (approved):

PR #400, opened 2019-02-18
Remove analytics consent dialog when the tab is not there

irlaec:


octoprint_mrbeam/static/js/wizard_analytics.js
```diff
@@ -5,16 +5,18 @@ $(function () {
...
             if (next !== self.MY_WIZARD_TAB_NAME && current === self.MY_WIZARD_TAB_NAME) {
                 let result = self._handleAnalyticsTabExit();
+                self.containsAnalyticsTab = true;
                 return result;
             }
         };
 
         self.onBeforeWizardFinish = function() {
-            if (!self.analyticsInitialConsent()) {
+            if (!self.analyticsInitialConsent() && self.containsAnalyticsTab) {
```
andywerner, line 19: If the analytics wizard is not shown, would onWizardFinish() still be called? What data would be sent to the server? I think, onWizardFinish() needs to have the self.containsAnalyticsTab as well.

andywerner (changes requested):

octoprint_mrbeam/static/js/wizard_analytics.js
irlaec: You're right. I've added it.

andywerner (approved):

PR #432, opened 2019-03-22
added frontend warning if grbl auto update was not successful

andywerner:
Screenshot 2019-03-22 17 40 40


lina21:
"./translations/de" is not synchronised with "./octoprint_mrbeam/translations/de"

PR #435, opened 2019-03-25
F international

lina21:
Removed "translations/de", which was the same as "octoprint_mrbeam/translations/de"


andywerner:
old po file. DO NOT MERGE

PR #447, opened 2019-04-08
Add laser power correction

irlaec:


irlaec:
[(4) Laser Power Correction](https://trello.com/c/jZcBPxLh/46-4-laser-power-correction)

octoprint_mrbeam/__init__.py
```diff
@@ -133,6 +133,13 @@ def initialize(self):
 
 		self.focusReminder = self._settings.get(['focusReminder'])
 
+		self.lh['serial'] = self._settings.get(["laserhead", "serial"])
```
andywerner: could you put this into a separate function that return the lh dict to have less clutter here in the initialize method. thx

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, plugin):
 		self._storedConversions = list()
 
 		self._jobevent_log_version = 4
-		self._deviceinfo_log_version = 4
+		self._deviceinfo_log_version = 5  # Changed in v61 (04-04-2019)
```
andywerner: v0.1.61

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -196,6 +196,12 @@ def __init__(self, port=None, baudrate=None, callbackObject=None, printerProfile
 		self._pluginManager = octoprint.plugin.plugin_manager()
 		self._serial_factory_hooks = self._pluginManager.get_hooks("octoprint.comm.transport.serial.factory")
 
+		# laser power correction
+		if _mrbeam_plugin_implementation._settings.get(['laserhead', 'correction', 'enabled']):
+			self._power_correction_factor = _mrbeam_plugin_implementation._settings.get(['laserhead', 'correction', 'factor'])
```
andywerner: why don't you use your method _set_power_correction_factor() here? Wouldn't it be nicer if you did?

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1340,35 +1351,6 @@ def _handle_command_handler_result(self, command, command_type, gcode, handler_r
...
-					self._current_feedrate = int(temp)
-				elif temp < 30:
-					temp = 30
-					self._current_feedrate = int(temp)
-				self.sendCommand('F%d' % self._current_feedrate)
-
-	def _set_intensity_override(self, value):
-		temp = value / 100.0
-		if temp >= 0:
-			self._intensity_factor = temp
```
andywerner, line 1362: can you please check if the following are still in use and remove them if not:
- self._feedrate_factor
- self._feedrate_dict
- self._actual_feedrate
- self._intensity_factor
- self._actual_intensity
- self._intensity_dict

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1398,19 +1380,22 @@ def _replace_intensity(self, cmd):
...
-				else:
-					new_intensity = int(round(self._actual_intensity * self._intensity_factor))
-					if new_intensity > intensity_limit:
-						new_intensity = intensity_limit
-					self._intensity_dict[intensity_cmd] = new_intensity
-				self._current_intensity = int(round(new_intensity))
-				return cmd.replace(intensity_cmd, 'S%d' % self._current_intensity)
+
+			# Limit GCode input
+			new_intensity = int(round(self._actual_intensity))
```
andywerner: Isn't this already done a few lines above?
I like your code more. But let's clean up the ole one. I might be wrong, but I still think that we don't need the self._actual_intensity any more.

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1398,19 +1380,22 @@ def _replace_intensity(self, cmd):
...
+			if new_intensity > intensity_limit:
+				new_intensity = intensity_limit
+
+			# Apply power correction factor and limit again
+			new_intensity = int(round(new_intensity * self._power_correction_factor))
+			if new_intensity > self.MAX_INTENSITY_AFTER_CORRECTION:
+				new_intensity = self.MAX_INTENSITY_AFTER_CORRECTION
+
+			self._logger.info('Intensity command changed from S{old} to S{new} (correction factor {factor})'.format(
```
andywerner: this would be logged with every GCode command. Please don't.
If you'd like to see this for debugging, you can keep the line and either comment it out or you introduce a global contant like DEBUG, this line gets only logger if DEBUG == True but you set to False before you commit.

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1603,6 +1588,39 @@ def _handle_user_command(self, cmd):
...
+
+					elif token == 0:
+						self._log("Disabling power correction...")
+						_mrbeam_plugin_implementation._settings.set_boolean(["laserhead", "correction", "enabled"], False, force=True)
+						_mrbeam_plugin_implementation._settings.save()
+						_mrbeam_plugin_implementation.lh['correction_enabled'] = False
+						self._power_correction_factor = 1
+				else:
+					self._log("No parameter given (0 or 1)")
+			elif specialcmd.startswith('/terminal_show_checksums'):
```
andywerner: please rename it to "show_checksums". we're already in the terminal...

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1603,6 +1588,39 @@ def _handle_user_command(self, cmd):
 			elif specialcmd.startswith('/correct_settings'):
 				self._log("Correcting GRBL settings...")
 				self.correct_grbl_settings()
+			elif specialcmd.startswith('/power_correction'):
```
andywerner, line 1596: Is it possible to set another correction factor here?
It's not. Just double checking.

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1613,6 +1631,8 @@ def _handle_user_command(self, cmd):
 				self._log("   /disconnect")
 				self._log("   /reset")
 				self._log("   /correct_settings")
+				self._log("   /power_correction <x>  		--> Enable: x = 1; Disable: x = 0")
+				self._log("   /terminal_show_checksums <x>  --> Enable: x = 1; Disable: x = 0")
```
andywerner: also rename pls

andywerner (changes requested):

andywerner:
One more thing, Iratxe: if we manually write a correction factor to config.yaml it would get overwritten immediately once a new value from iobeam is received, right?
Can you please introduce a key to config.yaml that is 'factor_override'. If it has a value, this value will be used as correction factor, regardless.

octoprint_mrbeam/__init__.py
irlaec: done!

octoprint_mrbeam/analytics/analytics_handler.py
irlaec: changed

octoprint_mrbeam/printing/comm_acc2.py
irlaec: I think that here we may not have the _mrbeam_plugin_implementation.lh initialized yet, does it make sense?

octoprint_mrbeam/printing/comm_acc2.py
irlaec: True, it's changed now

octoprint_mrbeam/printing/comm_acc2.py
irlaec: Deleted!

octoprint_mrbeam/printing/comm_acc2.py
irlaec: done

octoprint_mrbeam/printing/comm_acc2.py
irlaec: done

octoprint_mrbeam/printing/comm_acc2.py
irlaec: all of them are removed now

octoprint_mrbeam/__init__.py
```diff
@@ -171,6 +171,7 @@ def _initialize_lh(self):
 		self.lh['serial'] = self._settings.get(["laserhead", "serial"])
 		self.lh['correction_factor'] = self._settings.get(["laserhead", "correction", "factor"])
 		self.lh['correction_factor_override'] = self._settings.get(["laserhead", "correction", "factor_override"])
+		self.lh['gcode_intensity_limit'] = self._settings.get(["laserhead", "correction", "gcode_intensity_limit"])
```
andywerner:
```suggestion
```

andywerner (changes requested):

andywerner (approved):

PR #467, opened 2019-05-03
Fix #441: move design with keyboard arrows

### LINKED ISSUE #441: Can't move a design in the workspace with the arrows
If a design that is placed in the workspace is scaled with the mouse, and then you try to move it from the keyboard (with the arrows), it instead scales more and more.
To fix this it is necessary to click outside of the image and then click on the image again, this way you can move the image with the arrows again.

irlaec:


irlaec:
[Can't move a design in the workspace with the arrows](https://trello.com/c/O73Bz8d1/83-cant-move-a-design-in-the-workspace-with-the-arrows)

andywerner (approved):

PR #468, opened 2019-05-03
Quickshape name lookup

irlaec:


irlaec:
[Change filenames (affects job time estimation)](https://trello.com/c/tBo5IjLw/69-change-filenames-affects-job-time-estimation)

Issue #218, opened 2017-11-13
Time display (passed time) doesn't update during long running Gcode commands


Gallore:
Looks like while a long running Gcode is performed, the backend keeps sending the same time to the frontend, so that the time counter there stops counting. Once the Gcode command has finished (and the next one gets processed) the time gets updated to the then correct value.

<img width="366" alt="screenshot 2017-11-13 18 01 48" src="https://user-images.githubusercontent.com/4582388/32738362-ec2de084-c89c-11e7-9086-3c5f6c6b986b.png">

Find SVG and GCO files attached:
[issue#218.zip](https://github.com/mrbeam/MrBeamPlugin/files/1467707/issue.218.zip)

Gallore:
Maybe we simple handle time on the frontend. Just keep counting there...

irlaec:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix "Laser Time" display](https://trello.com/c/Pwx8Y4Q4/68-fix-laser-time-display)

PR #477, opened 2019-05-13
Fix laser time display

irlaec:


octoprint_mrbeam/static/js/mother_viewmodel.js
```diff
@@ -230,6 +234,8 @@ $(function () {
                  self.terminal.checkAutoscroll();
             });
             self.terminal.activeAllFilters();
+
+            self.state._processProgressData = function(){};
```
andywerner, line 241:
```suggestion
            // MR_BEAM_OCTOPRINT_PRIVATE_API_ACCESS
            // our implementation here should be used instead of octoprints
            // to fix issues with the laser job time display
            self.state._processProgressData = function(){};
```

octoprint_mrbeam/static/js/mother_viewmodel.js
```diff
@@ -401,14 +407,35 @@ $(function () {
             }
         };
 
+        self.state.isPaused.subscribe(function (newIsPaused) {
```
andywerner, line 413: just FYI:
you could also implement a method `self.onPrintPaused` which would have the same effect.

andywerner (changes requested):

andywerner (approved):

irlaec:
[Fix "Laser Time" display](https://trello.com/c/Pwx8Y4Q4/68-fix-laser-time-display)

irlaec:
Fixes #218

PR #504, opened 2019-06-11
Add MrBeamPlugin version to analytics header

irlaec:


irlaec:
Not sure if we should increase the analytics version number, and in the case we do, for which type of record.

andywerner (changes requested):
Don't we need to bump the version number of all types?

PR #506, opened 2019-06-13
Add http self check on startup

irlaec:


irlaec:
[MrBeamPlugin: http self check after start up](https://trello.com/c/cGsgWfOf/119-mrbeamplugin-http-self-check-after-start-up)

PR #544, opened 2019-07-29
Refactor wizard configuration code

irlaec:


octoprint_mrbeam/wizard_config.py
```diff
@@ -0,0 +1,149 @@
...
+class WizardConfig:
+	def __init__(self, plugin):
+		self._logger = mrb_logger("octoprint.plugins.mrbeam.wizard_config")
+
+		self._plugin = plugin
+		self._user_manager = plugin._user_manager
+		self._plugin_manager = plugin._plugin_manager
+		self._settings = plugin._settings
+
+		self._wizard_version = 16  # random number. but we can't go down anymore, just up.
```
andywerner: This should be a constant

octoprint_mrbeam/wizard_config.py
```diff
@@ -0,0 +1,149 @@
+from flask.ext.babel import gettext
+from octoprint_mrbeam.mrb_logger import mrb_logger
+
+
+class WizardConfig:
+	def __init__(self, plugin):
```
andywerner, line 6: Could you please add a comment that explains where the order of the wizard-tabs is defined. Just to help our future selfs. thx

octoprint_mrbeam/wizard_config.py
```diff
@@ -0,0 +1,149 @@
...
+				mandatory=False,
+				suffix='_analytics',
+				template='wizard/wizard_analytics.jinja2',
+				div='wizard_plugin_corewizard_analytics',
+			)
+		)
+
+		return welcome_wizard_tabs
+
+	@staticmethod
```
andywerner: Why is this one a static method while _welcome_wizard_config() is a regular member?
Does the call to this one even work with self._whatsnew_wizard_config() (line 30)

andywerner (changes requested):

octoprint_mrbeam/wizard_config.py
irlaec: Good idea

octoprint_mrbeam/wizard_config.py
irlaec: Changed

octoprint_mrbeam/wizard_config.py
irlaec: Done

PR #559, opened 2019-08-08
Check if lh serial is valid

irlaec:


irlaec:
[Check if laser head is valid](https://trello.com/c/gbjnVxQO/206-check-if-laser-head-is-valid)

andywerner (approved):

PR #588, opened 2019-09-19
Handling of nested SVG tags

andywerner:
…is way though


andywerner:
Unfortunately we're still not there. Now results are too big:

Small: Result when we simply remove all SVG elements, big with current svg handling.
Same when original is compared from Chrome (pure) and temp.svg from Mr Beam

Screenshot 2019-09-20 at 10 20 43

andywerner:
Closing this since we decided that it's not worth to keep working on it,

PR #594, opened 2019-10-01
Customer reviews

irlaec:


andywerner:
I know I said you should use the laser safety screen as a template… But now I found that there is an easier way to read user settings on the frontend from the LoginStateViewModel: (here accessed via workinAreaViewModel:)
`mrbeam.viewModels.workingAreaViewModel.loginState.currentUser()`

I’d like to discuss with you tomorrow if it makes sense to refactor your code for the reviews so that it takes advantage of this.
One potential pitfall of this above is that I don’t know when/if it updates during runtime.

andywerner:
Theres also the userSettingsViewModel:
https://docs.octoprint.org/en/master/plugins/viewmodels.html#dependencies

andywerner:
`mrbeam.viewModels.userSettingsViewModel.updateSettings(mrbeam.viewModels.workingAreaViewModel.loginState.currentUser().name, {mrbeam:{andytest:'hallo'}})`

this one wipes all mrbeam suer settings. So a logged in suer can add any settings he wants.

octoprint_mrbeam/__init__.py
```diff
@@ -292,6 +295,7 @@ def get_settings_defaults(self):
 			),
 			grbl_version_lastknown=None,
 			tour_auto_launch = False,
+			num_succ_jobs = 0,
```
andywerner: move to usage,yaml

octoprint_mrbeam/__init__.py
```diff
@@ -1249,6 +1259,16 @@ def on_api_command(self, command, data):
...
 
+	def get_user_name(self):
+		from flask.ext.login import current_user
+
+		# Looks like current_user sometimes does not work, so we save it and the next time if there's no information
+		# we just use the last saved user.
+		if current_user and not current_user.is_anonymous():
+			self._current_user = current_user.get_name()
+
+		return self._current_user
```
andywerner, line 1277: not needed any more

octoprint_mrbeam/static/js/review.js
```diff
@@ -0,0 +1,150 @@
...
+        self.justGaveReview = ko.observable(false);
+
+        self.shouldAskForReview = ko.computed(function(){
+            if (self.loginState.currentUser() && self.loginState.currentUser().active) {
+                let numSuccJobs = self.loginState.currentUser().settings.mrbeam.review.num_succ_jobs;
+                let reviewGiven = self.loginState.currentUser().settings.mrbeam.review.given;
+
+                console.log('############ NUM JOBS: '+numSuccJobs);
+                console.log('############ GIVEN: '+reviewGiven);
+                console.log('############ NOW YES: '+self.justGaveReview());
```
andywerner: remove

andywerner (changes requested):

octoprint_mrbeam/__init__.py
irlaec: I actually need it...

octoprint_mrbeam/__init__.py
irlaec: moved

octoprint_mrbeam/static/js/review.js
irlaec: removed

PR #605, opened 2019-10-11
Welcome Wizard: Error notification if user account could not be not c…

andywerner:
…reated.


andywerner:
Screenshot 2019-10-11 at 15 49 36

PR #624, opened 2019-10-25
introducing model_id with dreamcut ready

andywerner:


octoprint_mrbeam/__init__.py
```diff
@@ -1731,6 +1750,18 @@ def getSerialNum(self):
 			self._serial_num = self._get_val_from_device_info('serial')
 		return self._serial_num
 
+	def get_model_id(self):
+		"""
+		Gives you the device's model id liek Mrb2 or MrB2-DC
```
amrsoll:
```suggestion
		Gives you the device's model id like MRBEAM2 or MRBEAM2_DC_R1
```

irlaec (approved):

PR #627, opened 2019-10-30
 Get material settings depending on current device

amrsoll:


octoprint_mrbeam/__init__.py
```diff
@@ -1798,7 +1799,6 @@ def getHostname(self):
 				self._logger.warn("getHostname() system hostname got changed to: {}. Requires reboot to take effect!".format(hostname_dev_info))
 
 
-		return self._hostname
```
andywerner, line 1801: This looks like removed by accident! If now, pls explain.

andywerner (changes requested):

octoprint_mrbeam/__init__.py
amrsoll: sorry O:) thank you for pointing it out!!

PR #635, opened 2019-11-06
B highlight designs onmouseover

hungerpirat:
simpler css, simpler code. works for images, svgs, dxf, quick*


andywerner (approved):
Cancel Slincing improved but still not good enough. -> reverting UI

PR #649, opened 2019-11-17
renderPNG: crops img to bbox instead of rendering whole workingarea.

hungerpirat:
renderPNG / renderInfills now takes care about the used space in the working area.
Advantage: The backend (img2gcode.py) gets way less data and is therefore faster. 
Formerly the rendered backgrounds were at least 2500x1950, most of the time with default line distance of 0.15mm it was 3333x2600.
With this image sizes the necessary processing steps (pixel reduction, remove transparency, contrast, greyscale, sharpen, dithering, separation) take way less time and memory.


andywerner:
The background of the resulting png is now white, before it was transparent.

andywerner (approved):

PR #651, opened 2019-11-19
Crop analytics file when it becomes too big

irlaec:


irlaec:
[Analytics: prevent file from getting too big](https://trello.com/c/DaF4zHLv/301-analytics-prevent-file-from-getting-too-big)

andywerner (approved):

PR #655, opened 2019-11-26
Read power target value from laser head

irlaec:
There is a method we are not using at all: def _validate_lh_serial.
Should we remove it or use it somewhere?

### DIFF
--- a/octoprint_mrbeam/iobeam/laserhead_handler.py
+++ b/octoprint_mrbeam/iobeam/laserhead_handler.py
@@ -17,7 +17,7 @@ def laserheadHandler(plugin):
 
 
 class LaserheadHandler(object):
-	LASER_POWER_GOAL = 950
+	LASER_POWER_GOAL_DEFAULT = 950
 	LASERHEAD_SERIAL_REGEXP = re.compile("^[0-9a-f-]{36}$")
 
 	def __init__(self, plugin):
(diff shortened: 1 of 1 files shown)


irlaec:
[Add "target_power" to laser head data and send it to Plugin for power correction. (1)](https://trello.com/c/D88cYIol/329-add-targetpower-to-laser-head-data-and-send-it-to-plugin-for-power-correction-1)

octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -152,31 +153,36 @@ def _calculate_power_correction_factor(self):
...
 		p_75 = None
 		p_85 = None
+		target_power = self.LASER_POWER_GOAL_DEFAULT
 		if power_calibration:
 			p_65 = power_calibration.get('power_65', None)
 			p_75 = power_calibration.get('power_75', None)
 			p_85 = power_calibration.get('power_85', None)
+			target_power = power_calibration.get('target_power', self.LASER_POWER_GOAL_DEFAULT)
+
+			if target_power < 0 or target_power >= 1300:
```
andywerner: pls define the limit in a constant. 
If target_power > 130 pls set target_power = 1300

andywerner (changes requested):

octoprint_mrbeam/iobeam/laserhead_handler.py
irlaec: Changed

PR #657, opened 2019-11-26
Laser param screen: rewording, arrows and smaller details

irlaec:
This is how it looks like now: (I'm not 100% sure about how the arrows look, as they are no longer aligned with the min/max labels from the compressor)
image
Is the wording for speed/intensity good?


irlaec:
[Laser param screen (3)](https://trello.com/c/yLgbqfcE/342-laser-param-screen-3)

andywerner (approved):

PR #675, opened 2019-12-09
Bug fix: show notification to user when using an older iobeam

irlaec:


octoprint_mrbeam/iobeam/iobeam_handler.py
```diff
@@ -264,22 +263,32 @@ def _send_command(self, command):
...
+			return False
 		else:
-			return True, 0
-
-
-	# return LooseVersion(self.iobeam_version) >= LooseVersion(self.IOBEAM_MIN_REQUIRED_VERSION) and LooseVersion(self.iobeam_version) < LooseVersion(self.IOBEAM_JSON_PROTOCOL_VERSION)
+			return True
+
+	def notify_user_old_iobeam(self):
+		self._logger.exception(
```
andywerner: log on exception level? I think error would be adequat

andywerner:
```suggestion
		self._logger.error(
```

andywerner (approved):

PR #682, opened 2019-12-23
F loading overlay animation

hungerpirat:
Just some fun as I could not sleep before christmas.


andywerner:
see f_loading_animation_andy

PR #683, opened 2019-12-25
Show awesome focus reminder earlier; made checkbox label clickable.

hungerpirat:
show focus reminder directly after clicking laser...

### DIFF
--- a/octoprint_mrbeam/static/js/convert.js
+++ b/octoprint_mrbeam/static/js/convert.js
@@ -1214,6 +1214,26 @@ $(function(){
 		        self.sendFocusReminderChoiceToServer();
             }
         };
+		
+		self.move_laser_over_material = function() {
+			let x,y;
+			
+			// assumption: center of placed designs is over the material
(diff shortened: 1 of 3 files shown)


octoprint_mrbeam/static/js/convert.js
```diff
@@ -1197,6 +1214,26 @@ $(function(){
 		        self.sendFocusReminderChoiceToServer();
             }
         };
+
```
andywerner, line 1217:
```suggestion
/*
* not in use yet
*/
```

andywerner (approved):

andywerner:
Bugs:
- "it's focused" button is disabled (not int he screenshot below!) but should be enabled.
- If Laser focus reminder got enabled in settings, the focus reminder shows only after a reload of the frontend.

@hungerpirat : didn't you say that this "move over material" is not in use yet...?
Screenshot 2020-01-03 at 13 01 48

andywerner (changes requested):
see comment above

andywerner (approved):

PR #684, opened 2020-01-02
Add new detection algorthim

amrsoll:
- Single frame detection
- Auto adjust shutter speed to detect reference markers instead of the center of the frame

New python/pip packages needed:
- ~scipy (takes forever to install)~


amrsoll:
Still need to do before being semi-stable :

- Handle errors better (especially when start/stopping, opening lid etc...)
- get signal from front-end to know if it's ready to receive a refreshed picture
- Analytics : compare the results of the 2 algorithms
- Scale distortion matrix to accommodate higher resolution pictures
- Detect movement in order not to serve the same image twice

andywerner (changes requested):
we spoke about the todos

Marker detection ToDos:
- remember last know position of a marker (blocker for beta)
- slider: local maxima - check if it’s really a marker
- detected blobs: cycle is this a marker?
- reduce file size of the first image
- redo logic when markers flash on frontend
- Difference in file size between grey and colored.
- Timer to resend even with very slow connection
- use brightness for shutterspeed
- apply_best_shutter_speed() (maybe)

amrsoll:
- [x] remember last know position of a marker (blocker for beta)
- [ ] slider local maxima - check if it’s really a marker
- [ ] detected blobs: cycle is this a marker?
- [x] reduce file size of the first image
- [x] redo logic when markers flash on frontend
- [ ] use brightness for shutterspeed
- [ ] apply_best_shutter_speed() (maybe)
- [x] analytics documentation completed

Slow connection speeds (simulated regular 2G on Firefox):
First small picture served in 2 ~ 5 sec
Second bigger picture served in 14 sec

- [x] Difference in file size between grey and colored.

Low quality

- grayscale : 41 kB
- color : 45 kB

Better quality

- grayscale 418 kB
- color 492 kB


> - [x] Timer to resend even with very slow connection
> If front end hasn't shown sign that it is done downloading the last image, a new one will be sent.

andywerner (approved):

PR #709, opened 2020-01-16
Overshoot enabled. Added setting to advanced laser parameters to test overshoot

andywerner:
The overshoot feature was already implemented long time ago by Teja. This is just to make it accessible.

Once we release it publicly (STABLE) we should remove the frontend input field and set the overshoot_distance always to 1 mm.


hungerpirat (approved):
the naming "overshoot" is nothing official - just sth I had in mind. Maybe in the frontend is a better name for that.

PR #719, opened 2020-01-28
Add Desktop development support

amrsoll:
Free yourselves from the shackles of the ssh-development! Dev on your own machine, away from your MrBeam!

But more seriously, This is what @hungerpirat already does, and here we avoid having to install a fake picamera that ends up creating an unseen OSError instead of ImportError ;)

I'm still testing if everything is working as expected, and redacting a Wiki entry to set up a virtual environment with MrBeamPlugin and OctoPrint.


amrsoll:
pushed with changes from an other branch

PR #724, opened 2020-02-01
Leds Settings

hungerpirat:
Brightness and animation speed of the LED edges is now adjustable.


andywerner (approved):

andywerner:
In the settings screen a button "reset to defaults" or a visual indication, what the default values are would be good.

PR #738, opened 2020-02-13
Rm complex saturation and value dynamic changes

amrsoll:
They added complexity for no good reason.


amrsoll:
This also makes the detection algo much broader and versatile.
It doesn't seem to lose any performance with it

andywerner (approved):

PR #746, opened 2020-02-24
Calibration screen compatibility

amrsoll:
The calibration screen can now use the new algo for creating the pic_settings JSON.

NOTE : Both the new and the legacy algo use the same picture settngs. Given that their avg measurement is different, this means that the algo that is used to calibrate will give better results.

No extra effort was done in that regard given that:
- The new algo has priority when it comes to display a refreshed image
- We don't want to sit on a "One camera 2 systems policy" forever

NOTE2: It seems we get the "Requesting picture" message without visiting the calibration screen. Still investigating.

Bonus fix : The hostname in pic_settings.yaml is changed to reflect the device that last wrote into it

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -1536,10 +1536,16 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 		# transform dict
 		# todo replace/do better
 		newCorners = {}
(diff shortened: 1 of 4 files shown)


amrsoll:
Closing while figuring out some problems with x - y conversion

andywerner (approved):

PR #765, opened 2020-03-09
Adjust camera viewport to the material height

amrsoll:
Images produced by new algo can be dezoomed according to set object height

See left ribbon -> Preview -> Object height

Max allowed height : 38mm

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -1379,6 +1379,7 @@ class MrBeamPlugin(octoprint.plugin.SettingsPlugin,
 			material_settings=[],
 			on_camera_picture_transfer=[],
 			leds=[],
+			compensate_obj_height=[],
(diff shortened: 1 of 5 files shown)


amrsoll:
### Problem

I am convinced by the theoretical approach taken, but I have gotten mixed results. The higher you go, the more careful you have to be:
- The cardboard used to calibrate is not at Z=0 although it is interpreted as such
- If you are 1mm off when clicking on the arrowheads at Z ~ 2mm, you will be off by about 1mm horizontally for every 2 or 3 mm you distance yourself vertically. (i.e. you could be off by 1cm at 38mm height)

### Remediation

1. Laser a cardboard at both 0 and 38mm height, and use linear interpolation to keep the same precision at each height.
2. Increase camera resolution to the maximum to increase precision when clicking on the tip of the arrows.

Issue #773, opened 2020-03-12
Keep last HQ picture


amrsoll:
Given that shadows could trick the device into sending LQ pictures, it would be great to have a "HQ" only button.

amrsoll:
See #745

PR #775, opened 2020-03-12
Increase tolerance for motion detection.

amrsoll:
A lot of customers complain about excessive amounts of LQ pictures. Presumably, something might be moving on the workspace, probably just shadows.
This fix should make it more tolerable.


amrsoll:
Helps with #773 as the increased tolerance allows for light shadows to pass over the area without triggering a new image.

andywerner (approved):

PR #777, opened 2020-03-12
Start photo worker when interlocks open

amrsoll:
- Only take 1 Low qualiy picture before upscaling
- Start and stop photo worker faster


amrsoll:
Some users are confused when it takes a long time for an HQ picture to arrive, so I tuned it down to a single one.

andywerner (approved):

PR #782, opened 2020-03-13
Turn off Laserhead halfway through overshoot

### LINKED ISSUE #780: Overshoot creates visible marks on some materials
On some materials (leather) overshoot creates visible marks. (See image - gco attached as zip)
Current suggested workaround: set an pierce time of 1ms, this will disable overshoot.

However, we should come up with a full solution. This could wither be:
- add a switch to the advanced settings and add it to the material settings accordingly
- find a metric of how to turn it on/off automatically

IMG_20200311_213728
[basst_v1.1.gco.zip](https://github.com/mrbeam/MrBeamPlugin/files/4329793/basst_v1.1.gco.zip)

amrsoll:
Fixes #780


amrsoll:
Closed as an other overshoot tactic was found.

PR #783, opened 2020-03-13
Fix err when test true-ish values of np arrays.

amrsoll:
This is what happens when you rely on True-ish values :man_shrugging:


amrsoll:
Resolved in #720

PR #812, opened 2020-04-15
MMAL Error : Attempt to fix & Notify Front End it case it doesn't

amrsoll:
Attempt to completely avoid starting a PhotoCreator thread while an other one is using the camera (current belief behind the cause of the MMAL Error

This message shows in case the error happens
Screenshot from 2020-04-16 01-48-11

These are the threading.Event elements that makes it possible :

- `active` flag : whether the camera is in use
- `stopEvent` : whether we are trying to exit the current PhotoCreator thread

### DIFF
--- a/octoprint_mrbeam/static/js/user_notification_viewmodel.js
+++ b/octoprint_mrbeam/static/js/user_notification_viewmodel.js
@@ -46,7 +46,10 @@ $(function () {
                 text: gettext("The camera has had a small issue, please restart your Mr Beam if you need to use the camera to continue with your work."),
                 type: 'error',
                 hide: false,
-            }
+                knowledgebase: {
+                    url: 'https://support.mr-beam.org/en/support/solutions/articles/43000570474-error-camera-error',
+                }
+            },
         };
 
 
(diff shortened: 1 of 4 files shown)


amrsoll:
New error message displayed after 2 camera restarts : 
Screenshot from 2020-04-16 17-32-12

octoprint_mrbeam/static/js/user_notification_viewmodel.js
```diff
@@ -34,8 +34,23 @@ $(function () {
...
+                type: 'info',
+                hide: true,
+            },
+            err_cam_conn_err: {
+                title: gettext("Camera Error"),
+                text: gettext("The camera has had a small issue, please restart your Mr Beam if you need to use the camera to continue with your work."),
+                type: 'error',
+                hide: false,
+                knowledgebase: {
+                    url: 'https://support.mr-beam.org/en/support/solutions/articles/43000570474-error-camera-error',
```
andywerner:
```suggestion
                    url: 'https://support.mr-beam.org/support/solutions/articles/43000570474-error-camera-error',
```

andywerner (changes requested):

amrsoll:
TODO : Add analytics : Are we able to recover from this error?

andywerner (approved):

PR #815, opened 2020-04-17
Add Support for Firefox mouse scroll to zoom.

### LINKED ISSUE #98: Scroll issues on iPad
Let's collect all things around scrolling on iPas and all other touch devices

Needs to marked as scrollable:

- Software update progress output: ```#settings_plugin_softwareupdate_workingdialog_output```
- Conversion Dialog fully expanded

amrsoll:
Changes the wheel event .
According to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event) it is compatible everywhere.

Obviously one would need to disable the use of shift key for iOS, but if it's the case then we could change the status of #98

### DIFF
--- a/octoprint_mrbeam/static/js/working_area.js
+++ b/octoprint_mrbeam/static/js/working_area.js
@@ -2731,11 +2731,11 @@ $(function(){
 		};
 
 		self.wheel_zoom_monitor = function(target, ev){
-			var wheel = ev.originalEvent.wheelDelta;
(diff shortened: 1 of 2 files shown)


amrsoll:
N.B. `mouseDelta = -2 * deltaY` which means you might want to tweak the zoom factor.

andywerner (approved):

Issue #250, opened 2018-01-08
Camera Calibration: Waiting infinitely for undistorted image if device is LOCKED


Gallore:
This is a very rare case since the user has to laser the corner markings first. But if he reboots the device after this and does NOT home it, then there is no error message but there's also no image coming from the camera.

amrsoll:
Extra tags:
camera markers homing message warning picture calibration screen

PR #817, opened 2020-04-17
Calibration screen compatibility

amrsoll:
Dependent on #813

- Fix : Pictures load from the correct folder
- Move some dependence from camera_calibration.js to camera.js
- Fix : No longer loads a placeholder image continuously (calpic_wait.svg)
- Deprecate a certain pic settings key + fix saved settings.
- Freeze the picture we use when clicking the "Start Calibration" button
- Fix: Will first find the markers before checking if the pic settings is malformed (needed info for the calibration)
- Error msg when pic settings file is not present/malformed.
- The page no longer need to be refreshed to take the calibration into account.
- Fix Error where the working area ViewModel wouldn't load from the calibration screen
    
Edit : The photoworker is told to take a new pic when pic settings are refreshed.


amrsoll:
It looks like the analytics are broken when we are not on the root domain
>`analytics.js`
`AnalyticsViewModel.self.send_fontend_event`
POST http://mrbeam-8ae9.local/plugin/mrbeam/plugin/mrbeam/analytics 404 (NOT FOUND)

Overall that is a non-issue since it only happens on the Wateroot calibration screen.

amrsoll:
BTW, it also fixes #250

Issue #805, opened 2020-04-07
SVG issue: "display: none" attribute is ignored. Issue only visible in laser result!


Gallore:
Attached SVG file: only the lower part should be visible. the bunnies in the middle and top of the working area are set to `display: none` per class pre-defined style.

<img width="1088" alt="Screenshot 2020-04-07 at 18 57 45" src="https://user-images.githubusercontent.com/4582388/78698117-17897000-7902-11ea-8e31-c44dc817c8e9.png">

[Ostern.zip](https://github.com/mrbeam/MrBeamPlugin/files/4446012/Ostern.zip)

Gallore:
fixed

PR #818, opened 2020-04-18
B 805 display none

hungerpirat:
removes elements hidden by a reference to a css class setting display:none. 
removal is done after elements are placed, otherwise getComputedStyle does not return the correct value (as it hasn't been computed).


octoprint_mrbeam/static/js/working_area.js
```diff
@@ -678,6 +678,17 @@ $(function(){
 					'mb:origin': origin
 				});
 
+				// remove hidden elements with "display:none" via a css class (svg fragment needs to be placed to use getComputedStyle())
+				let allElems = newSvg.selectAll('*[class]');
+				console.log("found elements", allElems.length);
+				for (var i = 0; i < allElems.length; i++) {
+					var el = allElems[i];
+					if (window.getComputedStyle(el.node).display === 'none') {
```
andywerner, line 687: What about visibility and all sorts of opacity or colors with alpha value of zero?

octoprint_mrbeam/static/js/unref.js
```diff
@@ -143,5 +143,32 @@ Snap.plugin(function (Snap, Element, Paper, global) {
 			return src_elem;
 		}
 	};
+	
+	/**
+	 * dereferences all class="..." attributes and replaces them with style="..."
+	 *
+	 * @returns {undefined}
+	 */
+	Element.prototype.unref_classes = function (classDefs = null) {
```
andywerner: Is this method used anywhere? Looks like it's just a stub without functionality?

octoprint_mrbeam/static/js/working_area.js
andywerner: What if display:none is set per style attribute rather than class attribute?

andywerner (changes requested):

octoprint_mrbeam/static/js/working_area.js
andywerner, line 687:
```suggestion
					if (window.getComputedStyle(el.node).display === 'none') {
					// also check visibility:hidden
```

andywerner (changes requested):

octoprint_mrbeam/static/js/working_area.js
andywerner: all styles coming from style attr are handled somewhere else.

andywerner (approved):

andywerner:
fixed #805

PR #824, opened 2020-04-28
Added camera marker recognition to debug settings

andywerner:


amrsoll:
Used the setting in a simplified version. See #825

PR #838, opened 2020-05-12
Upload cam image to GCS per analytics

andywerner:


octoprint_mrbeam/__init__.py
```diff
@@ -1439,14 +1440,15 @@ def on_api_command(self, command, data):
...
 			# TODO ANDY Load materials when the user logs in as well
-
 			try:
 				return make_response(jsonify(parse_csv(laserhead=self.get_model_id())), 200)  # TODO : Give parse_csv the right laserhead type
 			except Exception as err:
 				self._logger.exception(err.message)
 				return make_response(err.message, 500)
 		elif command == "on_camera_picture_transfer":
 			self.lid_handler.on_front_end_pic_received()
+		elif command == "camera_caputre_for_analytics":
```
amrsoll: typo : camera_capture

octoprint_mrbeam/camera/undistort.py
```diff
@@ -148,6 +145,8 @@ def save_debug_img(img, name):
 
 	if debug_out or undistorted:
 		save_debug_img(img, "undistorted")
+	if analytics_callback:
+		analytics_callback(img)
```
amrsoll: I think it is best to not add the analytics here. I strived to keep the analytics to lidhandler.py.

+ This call is blocking. You probably want to either have it run after the client has been served (i.e. doing it from lidhandler.py) or run it in parallel.
+ You could also choose to compress the jpg image and then encode it in base64.
+ Alternatively, on my own branch, I moved the save method of the raw image outside of this function and into the lidhandler.

I don't think we gain much more information from the undistorted picture. Better to send the raw image + the distortion matrix - both of which can be accessed in lidhandler.py

octoprint_mrbeam/iobeam/lid_handler.py
```diff
@@ -297,50 +304,54 @@ def restart(self, pic_settings=None, cam_params=None, out_pic_size=None, blockin
...
+							notification_id='err_cam_conn_err',
+							replay=True))
+				return
+			except Exception as e:
+				if e.__class__.__name__.startswith('PiCamera'):
+					self._logger.exception("PiCamera_Error_while_preparing_camera_%s_%s", e.__class__.__name__, e)
+				else:
+					self._logger.exception("Exception_while_preparing_camera_%s_%s", e.__class__.__name__, e)
+			self.stopEvent.set()
+		except:
```
amrsoll, line 354: I don't understand what this extra try/except achieves. There is already one inside and this one doesn't log the exception.

octoprint_mrbeam/iobeam/lid_handler.py
```diff
@@ -553,7 +565,28 @@ def add_to_stat(pos, avg, std, mass):
 
 		except Exception as ex:
 			self._logger.exception('Exception_in-_save__s_for_analytics-_{}'.format(ex))
-
+	
+	def capture_last_img_for_analytics(self):
```
amrsoll: Name is misleading: It sounds like you are taking an extra picture for the pupose of the analytics.

If you want to upload a picture after the fact ( the client already received the picture, then decides to send it to the analytics ), then it would be best to send the latest image at the start of the serve_pictures function (for example at line 425 of this file) and not during the preparation of the image.

octoprint_mrbeam/static/js/camera.js
```diff
@@ -148,6 +150,10 @@ $(function(){
             }
             return result;
         };
+
+        self.capture_img_for_analytics = function(){
+            OctoPrint.simpleApiCommand("mrbeam", "camera_caputre_for_analytics", {})
```
amrsoll: same typo

amrsoll (changes requested):
So I don't think this is the right way to do it. I thought we decided to send files into a bucket instead of base64 strings to the analytics?

octoprint_mrbeam/iobeam/lid_handler.py
amrsoll: Or bypass the lidhandler completely by reading the cam/debug/undistorted.jpg file, encode and then send it.

PR #847, opened 2020-05-28
embed_gc(): Reverting modification from 4. Mach 2020 and adding a try/catch block

andywerner:


andywerner:
See #848 (SVG circle (and other) without ID: do better handling) on thoughts and potential improvements about this one.

PR #864, opened 2020-06-16
UI adjustments

andywerner:


andywerner:
Screenshot 2020-06-19 at 19 30 25

andywerner:
A few things I found: (no blockers!)

Shouldn't the text bloack be broader?
Screenshot 2020-06-19 at 19 32 29

New save material settings icons on the custom image are shifted a bit:
Screenshot 2020-06-19 at 19 33 20

User settings: a few buttons that are not round. (Best fix would be to hide the whole api key section....)
Screenshot 2020-06-19 at 19 34 40

For my taste the cancel button is too big. duscussable:
Screenshot 2020-06-19 at 19 34 22

PR #891, opened 2020-07-15
Fixes #882: Job Time estimations slowed down a bit. (With Analytics)

### LINKED ISSUE #882: Laser stops / goes very slow in the first seconds of a laser job
Reproduce:
1. Put a JPG/PNG on the working area, scale it big (that it covers ~ 1/4 of the working area)
2. Prepare the job. (I used cardboard single wave, engrave only, but I doubt that this matters)
3. Once the slicing is done start the job immediately. (Do not wait 30 secs or longer!!!)
4. During the first 30-60 sec the laser will stop or go very slow. After this time it goes back to normal.

If you re-try the same with the gcode, I dod not see the issue.

I assume that either it's the JobTimeEstimation or the testing of the rpm value of the fan during the first 30 sec which puts so much load on the cpu that it can't handle the streaming of the engraving gcode.

I tested on SW version 0.6.13.1 but I assume that this happens in PROD as well since the customer complained about it.

andywerner:


irlaec (approved):
Please add the new analytics to the documentation (if you didn't already) :)

PR #906, opened 2020-08-03
Save and load camera settings and markers across sessions and reboots

amrsoll:
Saves the settings each time the lid is closed or the user disconnects.


amrsoll:
And minor fixes (See commit messages)

andywerner (approved):

PR #908, opened 2020-08-04
User camera calibration (corner calibration)

irlaec:


octoprint_mrbeam/static/js/calibration/camera_calibration.js
```diff
@@ -748,10 +802,10 @@ $(function () {
 			});
 
 			self.resetView();
-			if(!window.mrbeam.isWatterottMode())
-				self.goto('#calibration_step_1');
 		};
 
+		// todo iratxe: should we abort the corner calibration as well?
+        // for now I call this when going back from the lens calibration...
 		self.abortCalibration = function () {
```
amrsoll: This function kept it's name from the time where there was only the corner calibration. It should either be called abortCornerCalibration or it should abort both calibration methods.

amrsoll (commented):
I also think that we are supposed to put all the pdf files into an other package, right @Gallore ?

octoprint_mrbeam/static/js/calibration/camera_calibration.js
irlaec: Sounds good! We don't have to change it for now as the lens calibration is still not done, probably something to work on in the next sprint.

PR #924, opened 2020-08-11
French translation.

amrsoll:


amrsoll:
See other translation merge

PR #935, opened 2020-08-13
iobeam analytics

irlaec:


irlaec:
[iobeam: send analytics data to OP about error rates](https://trello.com/c/aPV4Cqff/577-iobeam-send-analytics-data-to-op-about-error-rates)

andywerner (approved):

PR #938, opened 2020-08-14
Save imgs to tmp file before overwrite.

amrsoll:


octoprint_mrbeam/util/img.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
andywerner, line 1: why this?

andywerner (approved):

PR #945, opened 2020-08-18
Change camera status requirements and messages

irlaec:


andywerner:
Just tested:
- If the device is NOT homed but the lid is open, the Status shows all three checks red incl "Lid is closed". It should recognize that the lid is open and the first check should be green in this case.
- Once I homed, the first two checks switched to green as expected. The third one (markers) remained red for a while (as expected). But I had assumption that the text already said "All four markers found" but the color still showed red. Then it switched to green and the two camera images appeared. I'm not sure if this is even possible. Maybe it went too fast and I got confused....

andywerner:
Last issue:
- All checks green (as expected) but the text for the last check says: "No markers found since camera did not launch"
Reproduce: Have all checks be green, then close the lid: check 1 and 3 go to red and check three says "No markers found since camera did not launch", so far all expected. Then open the lid again: all checks go back to green (expected) but the text of check 3 does not change.

Screenshot 2020-08-18 at 11 56 17

andywerner (approved):

PR #947, opened 2020-08-18
Logging Overhaul

amrsoll:
The MrbLogger doesn't use a Logger anymore, because it is one. A lot has been delegated to the logging utilities themselves.

I do like the `force_kwargs` function I wrote if I do say so myself :P It allows to get rid of the hacky `kwargs['some_key'] = kwargs.get('some_key', default_value)` and averything is in the function definition instead, making it easier for the IDE and linters alike.

Anyway, this also allows us to use `logging.getLogger` or `mrb_logger.getLogger` interchangeably - they are essentially the same thing. The default Logger class has been set to `MrbLogger` as soon as the plugin is loaded in, and the `MrbLogger` is compatible with the standard use of Loggers


octoprint_mrbeam/mrb_logger.py
```diff
@@ -6,69 +6,83 @@
...
+			self._log(logging.ERROR, msg, args, analytics=analytics, **kwargs)
+
+	@force_kwargs(analytics=True)
+	def critical(self, msg, analytics=True, *args, **kwargs):
+		if self.isEnabledFor(logging.CRITICAL):
+			self._log(logging.CRITICAL, msg, args, analytics=analytics, **kwargs)
+
+	# Disable the exception stacktrace AXEL : but why?
+	@force_kwargs(analytics=True, exc_info=False,)
+	def exception(self, msg, analytics=True, exc_info=False, *args, **kwargs):
```
amrsoll, line 70: I kept that because of the original behaviour, but is it necessary?

octoprint_mrbeam/mrb_logger.py
```diff
@@ -78,67 +92,57 @@ def log(self, level, msg, *args, **kwargs):
...
 
-	def _terminal(self, level, msg, *args, **kwargs):
+	@force_kwargs(**DEFAULT_COMM)
+	def _terminal(self, level, msg, id=None, exc_info=False, *args, **kwargs):
 		global _printer
 
-		date = self._getDateString()
-		id = kwargs.pop('id', self.id_short)
+		date = MrbLogger._getDateString()
+		id = id or self.name_short # FIXME id is reserved
```
amrsoll, line 122: `id` should really be refactored to something else

octoprint_mrbeam/mrb_logger.py
```diff
@@ -152,30 +156,32 @@ def _analytics_log_event(self, level, msg, analytics_id, *args, **kwargs):
...
-				analytics_handler.add_logger_event(event_details, wait_for_terminal_dump=kwargs.get('terminal_dump', False))
+				analytics_handler.add_logger_event(
+					event_details, wait_for_terminal_dump=terminal_dump)
 
 			except:
-				self.logger.exception("Exception in _analytics_log_event: ")
+				self.exception("Exception in _analytics_log_event: ", analytics=False)
 		else:
-			self.logger.error('Could not write exception to analytics, the analytics handler was not initialized.')
+			self.error('Could not write exception to analytics, the analytics handler was not initialized.', analytics=False)
```
amrsoll, line 184: prevents loops in case the analytics failed to load

amrsoll:
I would like to merge this AFTER the Beta release, so I'm keeping it as a draft for now

amrsoll:
Latest update : Need to ask Gina about the plugin initialisation, and if the dirty MrBeamLogger is used everywhere.

PR #961, opened 2020-08-28
Fix: compressor slider was ignored fo any cutting job

andywerner:
For any cutting job, the value of the slider was ignored. Any cutting job had the compressor value set to 10. Engraving jobs were not affected.

### DIFF
--- a/octoprint_mrbeam/static/js/convert.js
+++ b/octoprint_mrbeam/static/js/convert.js
@@ -669,7 +669,7 @@ $(function(){
 				$(job).find('.param_passes').val(p.cut_p || 0);
 				$(job).find('.param_progressive').prop('checked', p.progressive);
 				$(job).find('.param_piercetime').val(p.cut_pierce || 0);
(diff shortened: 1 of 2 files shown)


octoprint_mrbeam/static/js/convert.js
```diff
@@ -669,7 +669,7 @@ $(function(){
 				$(job).find('.param_passes').val(p.cut_p || 0);
 				$(job).find('.param_progressive').prop('checked', p.progressive);
 				$(job).find('.param_piercetime').val(p.cut_pierce || 0);
-				$(job).find('.compressor_range').val(p.cut_compressor || 0);  // Here we pass the value of the range (0), not the real one (10%)
+				$(job).find('.param_cut_compressor').val(p.cut_compressor || 0);  // Here we pass the value of the range (0), not the real one (10%)
```
hungerpirat, line 672: Solution is fine. 
However I'd suggest to use the max value as a default instead of 0. This is more failsafe.

PR #972, opened 2020-09-02
Keep the camera turned on after user disconnects.

### LINKED ISSUE #960: Camera: stop disabling of the camera when a user disconnects
When a user disconnects, we stop the camera. The problem is that OP sometimes falsely assumes that the user is disconnected. We should remove the condition of a user being connected when we launch or stop the cam. 

_The alternative would be to add this requirement to the check-list in the camera status of the new camera page. and/or trigger a re-start of the camera through a button or when the camera page is opened._

amrsoll:
Fixes #960


amrsoll:
Fixes #960

andywerner (approved):

PR #973, opened 2020-09-02
Show the "Need Corner Calibration" msg on login.

### LINKED ISSUE #959: Warning message "Calibration needed" needs some changes
- Call it "Corner Calibration" - that's the new name and we ant to be precise in our error messages.
- The item in Settings menu is now called "Camera"
- The warning message appears only once all four markers are recognized. It should come when the camera starts (even if markers are hidden - this allows easier testing for Basti) or once the user logged in.
- (Optional) The status checklist in the camera settings should also show this warning. However if the corner calibration is done, it should not show a green check. It might be not worth the effort to implement this though.

Current state:
Screenshot 2020-08-25 at 15 21 08

amrsoll:
Fixes #959


amrsoll:
Any adjustments to camera_calibration.js will clash with #971

andywerner (approved):

PR #976, opened 2020-09-03
Fix image pre-processing sliders

irlaec:


amrsoll (approved):

amrsoll:
It also shows on the result :+1:

Issue #949, opened 2020-08-18
Bug: pierce time and progressive option do not show consistently if 2 or more cutting jobs


Gallore:
@khaledsherkawi Please add more details

Gallore:
![](https://github.trello.services/images/mini-trello-icon.png) [Bug: pierce time and progressive option do not show consistently if 2 or more cutting jobs](https://trello.com/c/NsNAWhDM/603-bug-pierce-time-and-progressive-option-do-not-show-consistently-if-2-or-more-cutting-jobs)

khaledsherkawi:
- Channel: Stable
- Affected version: 0.7.1 (and below)
- Component: Front-end
- Browser: Chrome
- Priority: Minor
- Description: While dealing with more than one cutting color in lasering and whenever you want to assign a different set of parameters to each color, some issues surface regarding the "advanced settings" parameters. These "advanced settings" parameters will be showing for one or more cutting colors and not for others in certain scenarios described below.
- Steps to reproduce:
1. GIVEN you added 3 shapes with 3 different line colors

Gallore:
![](https://github.trello.services/images/mini-trello-icon.png) [Bug: pierce time and progressive option do not show consistently if 2 or more cutting jobs](https://trello.com/c/XC4SfC7l/631-bug-pierce-time-and-progressive-option-do-not-show-consistently-if-2-or-more-cutting-jobs)

PR #990, opened 2020-09-14
Fix advanced options for multiple cut jobs

irlaec:


irlaec:
Fixes #949

irlaec:
[Bug: pierce time and progressive option do not show consistently if 2 or more cutting jobs](https://trello.com/c/XC4SfC7l/631-bug-pierce-time-and-progressive-option-do-not-show-consistently-if-2-or-more-cutting-jobs)

octoprint_mrbeam/static/js/convert.js
```diff
@@ -1707,6 +1707,10 @@ $(function(){
             return t_a - t_b;
         };
 
+		self._apply_binding_new_cut_jobs = function(newJob) {
```
andywerner, line 1710:
```suggestion
		self._apply_binding_new_cut_jobs = function(newJob) {
		// called from outside of the viewmodel, that's why self to reference VectorConversionViewModel is not available.
```

andywerner (approved):

PR #992, opened 2020-09-14
Add logging to ntp shift fix

irlaec:


octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -145,6 +145,12 @@ def _set_time(self, job_duration):
 				)
 				self._analytics_handler.add_job_ntp_sync_details(ntp_details)
 
+				self._logger.info('NTP shift fix - Job duration before: {}, after: {} --> shift: {}'.format(
+					ntp_details.get('time_shift'),
+					ntp_details.get('job_duration_before'),
+					ntp_details.get('job_duration_after'),
```
andywerner, line 151: does the last comma work here? it's not in a dict or a list. It's in the parameters of a function. ?

andywerner (approved):

octoprint_mrbeam/analytics/usage_handler.py
irlaec: I'll double check :+1:

PR #1001, opened 2020-09-17
Make code local server friendly.

### LINKED ISSUE #794: Remove mb_picture_calibration dependency
With  #793, we don't need this separate package anymore, so it can be deleted from the software update script.

amrsoll:
Change the pip updates slightly (see commit messages as well)
- Disable migration for X86 machines
- Fix stuff that wouldn't otherwise work on localhosts

Removes the legacy camera calibration as well (Fixes #794)
(Can be tested with mrbeam-8ae9 /dot/ local in the office)


octoprint_mrbeam/migrate.py
```diff
@@ -698,3 +704,15 @@ def set_lasercutterPorfile_2DEF(self, series):
                 profile_id,
                 default_profile,
             )
+
+
+def rm_camera_calibration_repo():
+    """Delete the legacy camera calibration and detection repo."""
+    from octoprint.settings import settings
```
andywerner: please log start end result

andywerner (approved):

Issue #873, opened 2020-07-02
Laser head stops moving when opening a big library


amrsoll:
**Problem**

When opening a very populated library, it has been observed that the laser head would jitter instead of moving smoothly during a laserjob because of bottle-necked gcode streaming.

**Recommandation(s) for resolution**

- Decrease the niceness of the GRBL streamer, or increase the niceness of all other tasks during a laserjob.
- Limit the number of files displayed simultaneously in the library.

An additional message could be shown in the library during a laserjob
>A laserjob is currently running. The design library will run slower than usual.

Gallore:
![](https://github.trello.services/images/mini-trello-icon.png) [Laser head stops moving when opening a big library](https://trello.com/c/lvjDFppA/633-laser-head-stops-moving-when-opening-a-big-library)

amrsoll:
Currently tasked with investigation slow Front End loading screen 

#### Device

OP v1.3.6  +  [MrBeamPlugin], netconnectd

Running on a RaspberryPi 3 with Ethernet connection

#### What we have experienced

The stutter has been mostly observed when loading up a crowded design library (via storage manager) when over the 150 files mark (I'm testing with ~200 files)   

Attached is the log and pstats files when running the code with a profiler after cleaning the metadata, loading the interface and letting the storage manager run in the background.

Loading the interface from scratch consistently takes ten's of seconds more than usual, and the OP doesn't seem to recognise the connection status of the user very well. 

#### Suspects

Given the age of this issue, it ain't easy - but it points me in the direction of limited I/O and the storage manager.

After some analysis, I have found that the storage manager with a large library is quite expensive between the use of yaml.dump and yaml.load (no news here), as well as dict_merge - used to update `.metadata.yaml` each time a file is processed. Granted, it's not quite as problematic when the `.metadata.yaml` has been built. 

#### Possible mitigations

Making the [GCode analysis tool sleep longer](https://docs.octoprint.org/en/1.3.12/configuration/config_yaml.html?highlight=profiler#gcode-analysis) should help ~ Doesn't help with the initial queue build-up (when starting OP).

I have written a small OP patch (attached) that allows dict_merge to run in place because of it's aforementioned presence on the profiler. - I've also written it for the latest OP/develop if interested. The in-place dict_merge calls were on average 2~3 times faster.

[octoprint_load_front_long.log](https://github.com/mrbeam/MrBeamPlugin/files/5264291/octoprint_load_front_long.log)
[...]

PR #1010, opened 2020-09-25
Filemanager optimisation

amrsoll:
Replace the GcodeAnalysisQueue and place a hook to future-proof the plugin for OP v1.4

The new Analysis Queue effectively does nothing. Each file take 0.04s to analyse

with 250 files in a library, that makes it 10 seconds, maybe 20 if we count some thread management overhead.

See #873 - This helps with slow interface loading speed when a gcode analysis runs in the background.

Can also greatly benefit from a faster dict_merge tool.


andywerner (approved):
add this marker for changes in OP

Issue #698, opened 2020-01-07
Translation values increment when exiting or entering a field


amrsoll:
It is extremely difficult to write a precise value in the 'translation' field of a design. It will almost always jump/round when exiting the field (e.g. pressing esc. when you want to apply the precise value)

It also happens when you modify other fields (size, rotation...) 

Super frustrating when you want to  align several designs precisely :O

hungerpirat:
there is a circular dependency between size, scale, position and rotation. 
additionally the calculated values are displayed rounded (see https://github.com/mrbeam/MrBeamPlugin/blob/develop/octoprint_mrbeam/static/js/working_area.js#L1043) 
this is probably the effect - to resolve the cause, the circular dependencies need to be removed.

PR #1024, opened 2020-10-05
Add Firefox as a supported browser.

amrsoll:
It gets tiring to see the warning when I no longer have any issues on Firefox.

We might want to restrict Firefox to the latest version.

The alternative would be to have a "Don't show this message again" checkbox.

Screenshot from 2020-10-05 11-19-23


amrsoll:
Nevermind, there is still issue #698 that doesn't give a good Firefox experience. Closing for now. But we should still have a check box to stop reminding us.

PR #1026, opened 2020-10-05
Add gcode tests.

amrsoll:
See the [test branch](https://github.com/mrbeam/MrBeamPlugin/tree/test) where this branch was already merged together with the rsc files for the tests.

@hungerpirat - you are welcome to tackle some of the tests yourself as you know the code like the back of your hand :wink: especially for various slicing modes or clipping over the area edges.

```
pip install pytest pytest-datafiles 
# if py3 env :
pip install cairosvg
# if py2: just have inkscape.
sudo apt install python-tk
# Mac: if python complains about tkinter : https://discourse.brew.sh/t/python3-installation-with-tkinter/3636/5 
cd path/to/MrBeamPlugin/
git checkout test
pytest # runs all the tests
```

Also, [install cairo](https://pycairo.readthedocs.io/en/latest/getting_started.html)

It produces a lot of output. You can configure the logging level from `pytest.ini`


octoprint_mrbeam/gcodegenerator/img_separator.py
```diff
@@ -340,9 +340,7 @@ def _get_contours(self, img, method=cv2.RETR_EXTERNAL):
             )
 
         else:
-            _, contours, hierarchy = cv2.findContours(
-                img, method, cv2.CHAIN_APPROX_SIMPLE
-            )
+            contours, hierarchy = cv2.findContours(img, method, cv2.CHAIN_APPROX_SIMPLE)
```
andywerner: Teja:

> das "_, " am anfang der Zeile ist wichtig, da opencv in verschiedenen versionen da unterschiedlich viele parameter zurück gibt. genau deshalb ist überhaupt diese versionsweiche da - und axel hat in seinem commit in beiden wegen den gleichen code.
```suggestion
            _, contours, hierarchy = cv2.findContours(img, method, cv2.CHAIN_APPROX_SIMPLE)
```

andywerner (changes requested):

octoprint_mrbeam/gcodegenerator/img_separator.py
andywerner: @amrsoll Please also check other OPenCV call, since OpenCV holds more fund thing like this. according to teja.

hungerpirat:
Added selenium based frontend testing.
Call: pytest --baseurl=http://mrbeam-7357.local --rsc_folder=/path/to/test_rsc-github-repo tests/frontend/
So far, the tests cover:
0. ui basic tasks: loading, login, etc...
1. critical designs (see https://github.com/mrbeam/test_rsc/tree/main/critical_designs)
2. quick shape feature

octoprint_mrbeam/gcodegenerator/img_separator.py
amrsoll: The underscore "variable" `_` just ignores the given returned element, but it is not a glob that catches multiple elements of an iterable.

```python
_, a = 1, 2, 3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: too many values to unpack (expected 2)
>>> _, a, b  = 1, 2
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 3, got 2)
```
The version of OpenCV that is commonly found on our devices only returns 2 elements, so either we:

- make a case-ish statement for different versions of OpenCV
- Freeze the OpenCV version

but we can't have it both ways

amrsoll: Also I believe that this change might be stale afte all of what Teja introduced :)

octoprint_mrbeam/gcodegenerator/img_separator.py
```diff
@@ -326,27 +327,29 @@ def _prepare_img_for_contour_separation(self, monochrome, threshold=255):
...
-            contours, hierarchy = cv2.findContours(img, method, cv2.CHAIN_APPROX_SIMPLE)
-            if isCV31:
-                self.log.info(
-                    "OpenCV "
-                    + cv2.__version__
-                    + " : filtering top level contours with img size cropped by one px on each side"
-                )
+            _, contours, hierarchy = cv2.findContours(
+                img, method, cv2.CHAIN_APPROX_SIMPLE
+            )
```
amrsoll: In OpenCV version 3.1 (the one used on our devices) it would seem that `findContours` only returns 2 variables.  I would recommend removing the extra mute `_,` because it is not flexible with the number of returned statements as explained in previous comment

octoprint_mrbeam/gcodegenerator/img_separator.py
hungerpirat: There is already a case-ish statement for different versions of openCV. Your removal of the _ in the second part broke the code @amrsoll

amrsoll: ah yes you are right I didn't see

amrsoll: Then fine, but shouldn't we be deleting references to older opencv code? Aren't all of the devices at least on cv2 version 3.4 ?

PR #1028, opened 2020-10-06
Limit laser param inputs + show in tooltip + refactor

### LINKED ISSUE #637: Laser params tooltip: Add min & max values for feedrate to tool tip text

irlaec:
Closes #637

### DIFF
--- a/octoprint_mrbeam/gcodegenerator/job_params.py
+++ b/octoprint_mrbeam/gcodegenerator/job_params.py
@@ -17,7 +17,7 @@ class JobParams:
         SPEED = 3000
         COMPRESSOR = 100
         PASSES = 10
-        PIERCE_TIME = 2000
+        PIERCE_TIME = 300
         LINE_DISTANCE = 1.0
         INTENSITY = 1300
 
--- a/octoprint_mrbeam/gcodegenerator/job_params.py
+++ b/octoprint_mrbeam/gcodegenerator/job_params.py
@@ -1,0 +1,30 @@
+class JobParams:
+    class Default:
+        INTENSITY_WHITE = 0
+        INTENSITY_BLACK = 50
+        FEEDRATE_WHITE = 1500
+        FEEDRATE_BLACK = 250
+        CONTRAST = 1.0
+        SHARPENING = 1.0
+        DITHERING = False
+        BEAM_DIAMETER = 0.15
+        PIERCE_TIME = 0
+        PIERCE_INTENSITY = 1000
+        ENG_COMPRESSOR = 100
+        PASSES = 1
+
+    class Max:
+        SPEED = 3000
+        COMPRESSOR = 100
+        PASSES = 10
+        PIERCE_TIME = 2000
+        LINE_DISTANCE = 1.0
+        INTENSITY = 1300
+
+    class Min:
+        SPEED = 50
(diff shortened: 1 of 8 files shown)


octoprint_mrbeam/static/js/convert.js
```diff
@@ -10,6 +10,37 @@ $(function () {
...
+                pierceTime: 0,
+                engCompressor: 0, // This is the value of the slider, different from the backend value
+                cutCompressor: 3,
+                passes: 1,
+            },
+            max: {
+                speed: 3000,
+                compressor: 3,
+                passes: 10,
+                pierceTime: 2000,
```
andywerner, line 32: set max to 300 ms

andywerner (approved):

octoprint_mrbeam/gcodegenerator/job_params.py
```diff
@@ -0,0 +1,30 @@
+class JobParams:
```
amrsoll, line 1: I would probably register this as a setting with `octoprint.settings.settings("mrbeam.job")` or so. It should then be possible to call all of it from the front end with `self.settingsViewModel.settings.plugins.mrbeam.job` or something like that.

I'm not quite sure it works, but in the worst case you can add this to the `self.on_default_settings` from `octoprint_mrbeam/__init__.py`, and that should work.

All of the leaves of the settings datastruct should be `ko.observables` so you don't have to declare everything 3 extra times in the code.

See https://github.com/mrbeam/MrBeamPlugin/pull/1031/commits/d5caf5bd91c96c0e4fbb2d632f39b2ff02de2f71#diff-f9f46dfe7b608ec38c4abe52ca160b75 for how I did it with `cam.remember_markers_across_sessions`

octoprint_mrbeam/static/js/convert.js
```diff
@@ -845,17 +878,26 @@ $(function () {
...
+            self.JOB_PARAMS.default.feedrateBlack
+        );
+        self.imgDithering = ko.observable(self.JOB_PARAMS.default.dithering);
+        self.beamDiameter = ko.observable(self.JOB_PARAMS.default.beamDiameter);
+        self.engravingPiercetime = ko.observable(
+            self.JOB_PARAMS.default.pierceTime
+        );
+        self.engravingCompressor = ko.observable(
+            self.JOB_PARAMS.default.engCompressor
+        );
```
amrsoll, line 900: refering to earlier, you can set the settings as such:
```
[...]
self.settingsViewmodel = params[1]
[...]

self.onBeforeBinding = function () {
    self.settings = self.settingsViewModel.settings.plugins.mrbeam.job;
};
```
And then use `self.setttings.default.intensityBlack()`

amrsoll (commented):
I think it could be improved using the OP settings manager :)

irlaec:
[Laser params tooltip: Add min & max values for feedrate to tool tip text #637](https://trello.com/c/McsdtsrE/606-laser-params-tooltip-add-min-max-values-for-feedrate-to-tool-tip-text-637)

amrsoll:
[Do the Settings Properly (from now on)](https://trello.com/c/Kk4ABsac/677-do-the-settings-properly-from-now-on)

PR #1032, opened 2020-10-08
 MrbFileManager: adding files to the design library

irlaec:


irlaec:
[Add a warning if user uploads a file with existing name. #862](https://trello.com/c/DE4MQD2u/604-add-a-warning-if-user-uploads-a-file-with-existing-name-862)

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -1 +1,72 @@
...
+
+        FileManager.__init__(
+            self,
+            self._plugin._analysis_queue,
+            self._plugin._slicing_manager,
+            self._plugin.laserCutterProfileManager,
+            initial_storage_managers=storage_managers,
+        )
+
+    def add_file_to_design_library(self, file_name, content, sanitize_name=False):
```
amrsoll, line 46: Wouldn't it have been easier to just stick with the regular [add_file](https://docs.octoprint.org/en/1.4.2/modules/filemanager.html#octoprint.filemanager.storage.LocalFileStorage.add_file)? At least for the file name or path - I get it for the content.

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -1 +1,72 @@
...
+        if sanitize_name:
+            file_name = self._sanitize_file_name(file_name)
+        content = self._sanitize_content(file_name, content)
+
+        file_obj = self.File(file_name, content)
+        self.add_file(
+            FileDestinations.LOCAL,
+            file_name,
+            file_obj,
```
amrsoll, line 55: file_obj should be in 2nd position?

amrsoll (commented):
Yes yes! I love it :D more octoprinty

octoprint_mrbeam/filemanager/__init__.py
irlaec: No:
`def add_file(self, destination, path, file_object, links=None, allow_overwrite=False, printer_profile=None, analysis=None, display=None)`

octoprint_mrbeam/filemanager/__init__.py
irlaec: What do you mean?

octoprint_mrbeam/filemanager/__init__.py
amrsoll: My bad! I was looking at LocalFileStorage

octoprint_mrbeam/filemanager/__init__.py
amrsoll: The file storage class already does some path cleanup. I don't exactly know why it has been split up in a different class, [it just seems to do more of what we want](https://github.com/mrbeam/OctoPrint/blob/ca3ba522305f83063445b0ead08aec6fd7c75c5b/src/octoprint/filemanager/storage.py#L739) :)

amrsoll: My double bad : the filemanager uses the storage add_file function. So in this case I presume everything is fine, perhaps we don't need to sanitise the name?

irlaec: The MrbFileManager is only used when we upload something from the code: after downloading a design from the ds, etc. For the normal user upload the file names are sanitized, because this is handled by octoprint, but for the design store case it crashed (with münchen_skyline for example). So I just use this function for this case

andywerner (approved):

PR #1040, opened 2020-10-14
DS menu item error fix

khaledsherkawi:
While in the Design store, whenever the "Design store" menu item is clicked, the design store should be showing the design listing instead of a single design page. On click, nothing happens and an error shows in the console.


khaledsherkawi:
[Design Store menu item bug](https://trello.com/c/IgZUhWwY/684-design-store-menu-item-bug)

PR #1041, opened 2020-10-14
Hide files after deletion (now it takes too long)

irlaec:
In the Mr Beams that have a lot of files, the deletion (both bulk and invididual) takes very long. So after deleting a design it doesn't disappear until some seconds after, and it feels like it didn't work.
Now the design will be hidden asap, even if the deletion process is not done yet.


irlaec:
[File deletion: remove files even before the deletion is made (for bulk deletion and for single deletion)](https://trello.com/c/eaJQgzbP/680-file-deletion-remove-files-even-before-the-deletion-is-made-for-bulk-deletion-and-for-single-deletion)

andywerner (approved):

PR #1055, opened 2020-10-30
Option to cancel final extraction

irlaec:


andywerner:
Let's test if the cancel `Air Filter run-on` button is a good choice. Could be confusing because the air filter does not stop once you click it, just switch to auto mode. It might be better to not have this button but to just cancel the run-on when the user opens the lid.

andywerner:
Laser-Job done animation is broken if run-on screen is shown.

PR #1077, opened 2020-11-17
Hard coded laser profiles

### LINKED ISSUE #1050: Insure the correct Laser cutter Profile is used.
What is the best option to make sure we have the correct profile in a new device? This includes new images that are used for repairs on older devices.

## Status Quo

The laser profiles can be found as yaml files in a folder under `.octoprint/printerprofiles/lasercutterprofiles/`. The plugin uses the name of the device found in `/etc/mrbeam` to chose the correct profile. If the device is a `MrBeam2U` model, and the `MrBeam2U.profile` is missing, then it only uses the `my_default.profile` - which can have the wrong settings. 

### Hard coded into the plugin (or other repo)

Not a bad strategy. It might be best to keep the configs in a separate repo for our sanity.

Advantages

- Easy to maintain / change / update
- Any overwriting settings file can partially overwrite the hard-coded defaults

Disadvantages

- CodeBase increase

### Based on the environment

Using the different available 

Advantages

- In theory - No need to tell the device which model it is. 

Disadvantages

- Not all model have a unique environment.
- Increased Prototyping complexity with behaviour depending on what is available for the machine.

### Add profiles to the update process

This is equivalent to adding a new repo containing the profiles. The setup script would put them into the correct folder on the device.

### GUI device configuration

In **Support mode**, one could :

- Change the device model and name
- Add changes on top of the default settings

This could be written as an additional OctoPrint plugin that is only enabled when in supprt mode.

Note: Some plugins **do not** require an OctoPrint restart to enable/disable.

amrsoll:
I just cannot understand why i am unable to laser any design, it doesn't make sense.

The 2U and 2C profiles are fine, and the `Profile` I deleted was not used. Re-instating it didn't help either.

Fixes #1050


octoprint_mrbeam/printing/profile.py
```diff
@@ -65,394 +73,252 @@ class InvalidProfileError(Exception):
...
+        #             lasercutterprofiles=dict(defaultProfile=LASER_PROFILE_DEFAULT)
+        #         ),
+        #     )
+        #     settings().save()
+        # else:
+        #     self._save_to_path(
+        #         self._get_profile_path(identifier),
+        #         profile,
+        #         allow_overwrite=allow_overwrite,
+        #     )
```
amrsoll: @Gallore I believe this is part of the deprecated feature where you could modify the profile from the front end?

octoprint_mrbeam/printing/profiles/mrb1.py
```diff
@@ -0,0 +1,34 @@
...
+    axes=dict(
+        x=dict(speed=5000, inverted=False),
+        y=dict(speed=5000, inverted=False),
+        z=dict(speed=1000, inverted=False),
+    ),
+    start_method=None,
+    grbl=dict(
+        resetOnConnect=False,
+    ),
+)
```
andywerner: remove Mr Beam 1 profiles. no need to support mrb1

octoprint_mrbeam/printing/profiles/dummy.py
```diff
@@ -0,0 +1,10 @@
+# Dummy laser config for the virtual Mr beam laser cutter
+
+__all__ = ["profile"]
+
+profile = dict(
+    id="MrBeam2Dummy",
+    name="Dummy Laser",
+    model="X",
+    start_method="onebutton",  # change ?
```
andywerner: start_method can be removed

octoprint_mrbeam/__init__.py
```diff
@@ -1515,8 +1501,12 @@ def engraveCalibrationMarkers(self, intensity, feedrate):
     # Laser cutter profiles
     @octoprint.plugin.BlueprintPlugin.route("/profiles", methods=["GET"])
```
andywerner, line 1502: please remove all everything that was meant to change profile from frontend. There's also a jinja file....

octoprint_mrbeam/printing/profile.py
```diff
@@ -65,394 +73,252 @@ class InvalidProfileError(Exception):
...
+            os.makedirs(_laser_cutter_profile_folder)
+        PrinterProfileManager.__init__(self)
+        self._folder = _laser_cutter_profile_folder
+        self._logger = mrb_logger(__name__)
+        # HACK - select the default profile.
+        # See self.select() - waiting for upstream fix
+        self.select(profile_id or settings().get(self.SETTINGS_PATH_PROFILE_DEFAULT_ID))
+
+    def _migrate_old_default_profile(self):
+        # TODO
```
andywerner: remove TODO comment

octoprint_mrbeam/printing/profile.py
```diff
@@ -65,394 +73,252 @@ class InvalidProfileError(Exception):
...
+        pass
```
amrsoll, line 133: OP related : migration script related - not used in our codebase

octoprint_mrbeam/printing/profile.py
```diff
@@ -65,394 +73,252 @@ class InvalidProfileError(Exception):
...
+
+    def _verify_default_available(self):
+        # Overloaded from OP because of printerProfiles path ``default_id``
+        default_id = settings().get(self.SETTINGS_PATH_PROFILE_DEFAULT_ID)
+        if default_id is None:
+            default_id = "_default"
+
```
amrsoll: Check : Are we still relying on the lasercutterSettings ?

octoprint_mrbeam/printing/profile.py
andywerner: please remove all code related to the feature to change profile from the frontend.

octoprint_mrbeam/printing/profile.py
```diff
@@ -65,394 +73,252 @@ class InvalidProfileError(Exception):
...
+    # @logme(output=True)
     def get_current_or_default(self):
-        if self._current is not None:
-            return self._current
-        else:
-            return self.get_default()
-
-    def get_current(self):
-        return self._current
+        return PrinterProfileManager.get_current_or_default(self)
```
amrsoll, line 234: Delete because not logging

octoprint_mrbeam/templates/settings/lasercutterprofiles_settings.jinja2
```diff
@@ -17,7 +17,7 @@
         <td class="settings_printerProfiles_profiles_model" data-bind="text: model, click: function() { $root.makeDefault($data); }" style="cursor:pointer"></td>
         <td class="settings_printerProfiles_profiles_action">
 <!--            <a href="#" class="icon-star" title="Select profile" data-bind="click: function() { $root.makeDefault($data); }"></a>&nbsp;|&nbsp;-->
-			<a href="#" class="icon-pencil" title="Edit Profile" data-bind="css: {disabled:$root.isReadOnly($data)}, click: function() { if(!$root.isReadOnly($data)) $root.showEditProfileDialog($data); }"></a>&nbsp;|&nbsp;<a href="#" class="icon-trash" title="Delete Profile" data-bind="css: {disabled:$root.isReadOnly($data)}, click: function() { if(!$root.isReadOnly($data)) $root.removeProfile($data); }"></a>
+			<a href="#" class="icon-pencil" title="Edit Profile" data-bind="click: function() { if(!$root.isReadOnly($data)) $root.showEditProfileDialog($data); }"></a>&nbsp;|&nbsp;<a href="#" class="icon-trash" title="Delete Profile" data-bind="css: {disabled:$root.isReadOnly($data)}, click: function() { if(!$root.isReadOnly($data)) $root.removeProfile($data); }"></a>
```
andywerner: please remove this file

andywerner (changes requested):

octoprint_mrbeam/printing/profile.py
amrsoll: but necessary

andywerner (approved):
We already talked about this one before. Not re-reading

PR #1084, opened 2020-11-19
Logging issues during convertion

andywerner:


octoprint_mrbeam/__init__.py
```diff
@@ -1748,162 +1748,174 @@ def gcodeConvertCommand(self):
...
 
-            r = make_response(jsonify(result), 202)
-            r.headers["Location"] = location
-            return r
+                r = make_response(jsonify(result), 202)
+                r.headers["Location"] = location
+                return r
+            except Exception as e:
+                self._logger.exception('Exception in gcodeConvertCommand() "/convert":')
+                raise e
```
amrsoll: Use the `@logExceptions` decorator instead. It also avoids lengthy diffs because of indentation and logs the function name.

octoprint_mrbeam/__init__.py
```diff
@@ -1748,162 +1748,174 @@ def gcodeConvertCommand(self):
...
+                            )
+                            printer.select_file(filenameToSelect, sd, True)
+
+                        # keep only x recent files in job history and gcode.
+                        self.mrb_file_manager.delete_old_files()
+                    except Exception as e:
+                        mrb_logger("octoprint.plugins.mrbeam").exception(
+                            "Exception in slicing_done(), callback of /convert call:"
+                        )
+                        raise e
```
amrsoll: `@logExceptions`

octoprint_mrbeam/__init__.py
```diff
@@ -1748,162 +1748,174 @@ def gcodeConvertCommand(self):
...
+                    select_after_slicing,
+                    print_after_slicing,
+                    append_these_files,
+                ):
+                    try:
+                        # append additional gcodes
+                        output_path = self.mrb_file_manager.path_on_disk(
+                            FileDestinations.LOCAL, gcode_name
+                        )
+                        with open(output_path, "ab") as wfd:
```
amrsoll: wfd? wtf?

octoprint_mrbeam/__init__.py
```diff
@@ -1748,162 +1748,174 @@ def gcodeConvertCommand(self):
...
+                            for f in append_these_files:
```
amrsoll: `f` is usually used for file descriptors. Confusing.

octoprint_mrbeam/__init__.py
```diff
@@ -1748,162 +1748,174 @@ def gcodeConvertCommand(self):
         del data["gcodeFilesToAppend"]
 
         if command == "convert":
```
amrsoll, line 1752: Shouln't the content of this `if` statement be isolated with the rest of the slicer files?

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -50,28 +55,33 @@ def __init__(self, plugin):
...
+                file_name,
+                file_obj,
+                links=None,
+                allow_overwrite=True,
+            )
+        except Exception as e:
+            self._logger_mrb.exception(
+                "Exception in MrbFileManager.add_file_to_design_library() ", test=True
+            )
+            raise e
```
amrsoll: `@logExceptions`

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -33,7 +38,7 @@ def save(self, absolute_dest_path):
 
     def __init__(self, plugin):
         self._plugin = plugin
-        self._logger = plugin._logger
+        self._logger_mrb = mrb_logger("octoprint.plugins.mrbeam.filemanager")
```
amrsoll, line 43: Why change from the standard `self._logger` ?

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -108,18 +118,24 @@ def _delete_files_by_age(self, files, num_files_to_keep=0):
...
     @staticmethod
     def _sanitize_content(file_name, content):
         _, extension = os.path.splitext(file_name)
-        if extension in (".svg", ".mrb", ".g", ".gco", ".gc", ".gcode", ".nc"):
+        extension = extension[1:].lower()
+        if (
+            extension
+            in MrbFileManager.FILE_EXTENSIONS_SVG
+            + MrbFileManager.FILE_EXTENSIONS_GCODE
+            + MrbFileManager.FILE_EXTENSIONS_HISTORY
```
amrsoll: For the sake of this, it could be nice to have `MrbFileManager.FILE_EXTENSIONS_ALL`

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -50,28 +55,33 @@ def __init__(self, plugin):
...
-
-        if self._settings.get(["gcodeAutoDeletion"]):
-            self.delete_old_gcode_files()
+        try:
+            self.delete_old_history_files()
+            if self._settings.get(["gcodeAutoDeletion"]):
+                self.delete_old_gcode_files()
+        except Exception as e:
+            self._logger_mrb.exception("Exception in delete_old_files()")
+            raise e
```
amrsoll: `@logExceptions`

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -21,6 +22,10 @@ class MrbFileManager(FileManager):
     MAX_HISTORY_FILES = 25  # TODO fetch from settings
     MAX_GCODE_FILES = 25  # TODO fetch from settings
 
+    FILE_EXTENSIONS_SVG = ["svg"]
+    FILE_EXTENSIONS_HISTORY = ["mrb"]
+    FILE_EXTENSIONS_GCODE = ["g", "gc", "gco", "gcode", "nc"]
+
```
amrsoll, line 30: I think the more octoprinty way would be to declare those file extensions in a hook. Well, that is, when we have our slicer code wrapped in a nice hook for OP 1.4
See https://docs.octoprint.org/en/master/plugins/hooks.html?highlight=filemanager#octoprint-filemanager-extension-tree

amrsoll (changes requested):
I will review it again once most of the try/excepts are gone :)

octoprint_mrbeam/filemanager/__init__.py
amrsoll: It can trigger different slicers depending on the files ( if I remember correctly )

octoprint_mrbeam/filemanager/__init__.py
andywerner: no, because we do want to exclude JPG and PNG

octoprint_mrbeam/filemanager/__init__.py
andywerner: this class is inhertited from OP. Either my mrb_logger is overwritten by OP or OPs is overwritten with mine. Intentionally used two different ones.

octoprint_mrbeam/__init__.py
```diff
@@ -1863,6 +1868,12 @@ def slicing_done(
                 # keep only x recent files in job history and gcode.
                 self.mrb_file_manager.delete_old_files()
 
+            # except Exception as e:
+            #     mrb_logger("octoprint.plugins.mrbeam").exception(
+            #         "Exception in slicing_done(), callback of /convert call:"
+            #     )
+            #     raise e
+
```
amrsoll, line 1876: Please delete the comments though

octoprint_mrbeam/mrb_logger.py
```diff
@@ -158,12 +164,23 @@ def _analytics_log_event(self, level, msg, analytics_id, *args, **kwargs):
...
+                    any(caller.filename.endswith(x) for x in REMOVE_FROM_STACKTRACE)
+                    or MRBEAM_PLUGIN_SEARCH_STR not in caller.filename
+                ):
+                    try:
+                        caller = getframeinfo(stack()[i][0])
+                        self.logger.info(
+                            "ANDYTEST i: %s, filename: %s",
+                            i,
+                            caller.filename if caller else "noCallerObj",
+                        )
```
amrsoll, line 180: Not sure if you are done yet - just seems like you are adding loglines without removing stuff from the stack

amrsoll: I think it would be good to have the `octoprint` package part of the stack in the stacktrace as well.

octoprint_mrbeam/mrb_logger.py
```diff
@@ -158,12 +164,23 @@ def _analytics_log_event(self, level, msg, analytics_id, *args, **kwargs):
         if analytics_handler is not None:
             try:
                 msg = msg % args if args and msg else msg
-                caller = None
-                caller_myself = getframeinfo(stack()[0][0])
+                caller = getframeinfo(stack()[0][0])
+                caller_myself = caller
                 i = 1
```
amrsoll, line 169: It would be best to remove the use of an index / counter by using a for / if loop:
```
for caller in ( getframeinfo(stack_elm[0]) for stack_elm in stack() ) :
    try:
        if ( _while_condition_ ) :
            # do stuff
    except:
        continue
```

octoprint_mrbeam/mrb_logger.py
```diff
@@ -158,12 +164,23 @@ def _analytics_log_event(self, level, msg, analytics_id, *args, **kwargs):
...
+                while (
+                    any(caller.filename.endswith(x) for x in REMOVE_FROM_STACKTRACE)
+                    or MRBEAM_PLUGIN_SEARCH_STR not in caller.filename
+                ):
+                    try:
+                        caller = getframeinfo(stack()[i][0])
+                        self.logger.info(
+                            "ANDYTEST i: %s, filename: %s",
+                            i,
+                            caller.filename if caller else "noCallerObj",
```
amrsoll, line 179: You have already insured `caller` is not `None` because your while loop uses `caller.filename` - ergo no need for 
`caller.filename if caller else "noCallerObj",` , just `caller.filename,`

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -108,18 +112,24 @@ def _delete_files_by_age(self, files, num_files_to_keep=0):
...
+        extension = extension[1:].lower()
+        if (
+            extension
+            in MrbFileManager.FILE_EXTENSIONS_SVG
+            # + MrbFileManager.FILE_EXTENSIONS_GCODE
+            # + MrbFileManager.FILE_EXTENSIONS_HISTORY
+        ):
             # TODO stripping non-ascii is a hack - svg contains lots of non-ascii in <text> tags. Fix this!
             content = "".join(i for i in content if ord(i) < 128)
         return content
```
amrsoll, line 135: I know it's no the purpose of the PR, but shouldn't the string be converted to a unicode string instead? Or replace the characters with something else so it's easier to diagnose (such as a "?").

I would mark this with a comment for the py3 transition - with unicode support backend into the language, this step would be redundant. You can also use `sys.version_info < (3, 0, 0)` to skip this step on python3.

amrsoll (changes requested):

octoprint_mrbeam/mrb_logger.py
amrsoll: Would also be nice to have the "in" and "out" nodes of the stacktrace displayed

amrsoll:
I should be possible to change `logException` to ignore itself by following advice from here:
https://stackoverflow.com/a/13898994/11136955

PR #1096, opened 2020-12-06
Support mode: removed 24hrs check and added onUserLoggedin check

andywerner:
removed 24hrs check on support file or calibration tool file because it drove Basti and Marcel nuts:
If there was an ntp time fix, the support file usually appeared older than 24hrs becasue it was created before the time fix happened.


andywerner:
we got a better idea

PR #1097, opened 2020-12-08
Support mode: removed 24hrs check, now checking number of usb sticks …

andywerner:
…mounted

### DIFF
--- a/octoprint_mrbeam/support.py
+++ b/octoprint_mrbeam/support.py
@@ -5,7 +5,6 @@ from octoprint_mrbeam.mrb_logger import mrb_logger
 
 SUPPORT_STICK_FILE_PATH = "/home/pi/usb_mount/support"
 CALIBRATION_STICK_FILE_PATH = "/home/pi/usb_mount/calibration_tool"
-SUPPORT_STICK_FILE_MAX_AGE = 60 * 60 * 24
 
 
 USER_NAME = "support@mr-beam.org"
@@ -29,7 +28,7 @@ def check_support_mode(plugin):
     try:
         if plugin._settings.get(["dev", "support_mode"]) or (
             os.path.isfile(SUPPORT_STICK_FILE_PATH)
-            and len(os.listdir(USB_MOUNT_DIR)) > 0  # if i a usb stick is plugged in
+            and len(os.listdir(USB_MOUNT_DIR)) > 0  # test if an usb stick is plugged in
         ):
             support_mode_enabled = True
         else:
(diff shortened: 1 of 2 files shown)


octoprint_mrbeam/support.py
```diff
@@ -13,6 +13,8 @@
 
 _logger = mrb_logger("octoprint.plugins.mrbeam.support")
 
+USB_MOUNT_DIR = "/media/"
+
```
amrsoll, line 16: Remove `SUPPORT_STICK_FILE_MAX_AGE`

octoprint_mrbeam/support.py
```diff
@@ -21,13 +23,14 @@ def check_support_mode(plugin):
...
     support_mode_enabled = False
+    if not hasattr(plugin, "_settings") or not hasattr(plugin, "_user_manager"):
+        return None
+
     try:
         if plugin._settings.get(["dev", "support_mode"]) or (
             os.path.isfile(SUPPORT_STICK_FILE_PATH)
-            and time.time() - os.path.getmtime(SUPPORT_STICK_FILE_PATH)
-            < SUPPORT_STICK_FILE_MAX_AGE
+            and len(os.listdir(USB_MOUNT_DIR)) > 0  # if i a usb stick is plugged in
```
amrsoll: Use [`os.path.isdir()`](https://docs.python.org/3/library/os.path.html#os.path.isdir)

octoprint_mrbeam/support.py
```diff
@@ -46,13 +49,14 @@ def check_calibration_tool_mode(plugin):
...
     mode_enabled = False
+    if not hasattr(plugin, "_settings") or not hasattr(plugin, "_user_manager"):
+        return None
+
     try:
         if plugin._settings.get(["dev", "calibration_tool_mode"]) or (
             os.path.isfile(CALIBRATION_STICK_FILE_PATH)
-            and time.time() - os.path.getmtime(CALIBRATION_STICK_FILE_PATH)
-            < SUPPORT_STICK_FILE_MAX_AGE
+            and len(os.listdir(USB_MOUNT_DIR)) > 0  # if i a usb stick is plugged in
```
amrsoll: Use [`os.path.isdir()`](https://docs.python.org/3/library/os.path.html#os.path.isdir)

amrsoll (changes requested):

octoprint_mrbeam/support.py
amrsoll: well it didn't show in the refered lines here, but you should delete the definition

octoprint_mrbeam/support.py
andywerner: Nope. I want the dirs in there. The number of dirs must be 1 or more!

octoprint_mrbeam/support.py
andywerner: same

octoprint_mrbeam/support.py
andywerner: No. I want to test if the number of subdirs in there is at least 1.

amrsoll (approved):

PR #1099, opened 2020-12-09
Implemented performant bulk deletion of gco or design files.

andywerner:


octoprint_mrbeam/__init__.py
```diff
@@ -2097,7 +2099,23 @@ def set_gcode_deletion(self, enable_deletion):
...
+            ):
+                self._logger.info(
+                    "set_gcode_deletion: Starting threaded bulk deletion of gcode files."
+                )
+                self._gcode_deletion_thread = threading.Thread(
+                    target=self.mrb_file_manager.delete_old_gcode_files,
+                    name="gcode_deletion_thread",
+                )
+                self._gcode_deletion_thread.daemon = True
+                self._gcode_deletion_thread.start()
```
amrsoll: Can be replaced like this:

```python
from octoprint_mrbeam.util import get_thread

self._gcode_deletion_thread = get_thread(daemon=True) (self.mrb_file_manager.delete_old_gcode_files) ()
```

See `util.__init__.py` line 82

octoprint_mrbeam/filemanager/__init__.py
```diff
@@ -106,12 +107,23 @@ def _delete_files_by_age(self, files, num_files_to_keep=0):
...
                 )
                 removals.append(tpl)
-
             sorted_by_age = sorted(removals, key=lambda tpl: tpl[0])
 
-            # TODO each deletion causes a filemanager push update -> slow.
+            files_to_delete = []
             for f in sorted_by_age[:-num_files_to_keep]:
-                self.remove_file(FileDestinations.LOCAL, f[1])
+                path = f[1]
```
amrsoll: Would be more transparent using 

```python
for _, path in sorted_by_age[:-num_files_to_keep]: # can also add as many _ as necessary
    # continue for loop as normal
```

octoprint_mrbeam/filemanager/file_storage.py
```diff
@@ -0,0 +1,90 @@
...
+from octoprint_mrbeam.mrb_logger import mrb_logger
+
+
+class MrBeamFileStorage(LocalFileStorage):
+    def __init__(self, basefolder, create=False):
+        self._logger_mrb = mrb_logger(
+            "octoprint.plugins.mrbeam.filemanager.mrbFileStorage"
+        )
+
+        LocalFileStorage.__init__(self, basefolder, create)
```
amrsoll: Revert order otherwise the parent `__init__` might overwrite it.

octoprint_mrbeam/filemanager/file_storage.py
```diff
@@ -0,0 +1,90 @@
...
+            try:
+                os.remove(file_path)
+                f_count += 1
+            except Exception as e:
+                self._logger_mrb.error(
+                    "Could not delete {name} in {path}, cause: {cause}".format(
+                        name=name, path=path, cause=e
+                    )
+                )
+                continue
```
amrsoll, line 47: Prefer `OSError` and `FileNotFoundError`:

```python
except OSError, FileNotFoundError:
    pass
```

octoprint_mrbeam/filemanager/file_storage.py
```diff
@@ -0,0 +1,90 @@
...
+    def _remove_metadata_multiple_entry(self, entries):
+        for path in entries.keys():
+            with self._get_metadata_lock(path):
+                metadata = self._get_metadata(path)
+
+                for name in entries[path]:
+                    if not name in metadata:
+                        continue
+
+                    if "hash" in metadata[name]:
```
amrsoll, line 76:
```py
from octoprint_mrbeam.util import dict_get
hash = dict_get(metadata, [name, "hash"])
if hash: # or is hash is not None:
    # do stuff
```

octoprint_mrbeam/__init__.py
amrsoll: Also, it the thread necessary now that it's quick?

amrsoll (changes requested):

octoprint_mrbeam/__init__.py
andywerner: nice. And yes, thread is always a good thing to respond faster to the frontend if possible

octoprint_mrbeam/filemanager/file_storage.py
amrsoll: oh nvm it's a different attribute

octoprint_mrbeam/filemanager/file_storage.py
andywerner: hmm... not sure if I agree. Anyway, this is copy-paste code from OP. Won't change

octoprint_mrbeam/filemanager/file_storage.py
andywerner: OP copy-past code.

amrsoll: Then please document the function:
- declare it's a copy-pasta from function x
- why it is a copy-pasta

amrsoll: Makes it easier to understand later on and to create a PR for OctoPrint

andywerner:
```INFO - remove_multiple_files(): Deleted 447/447 files in 1 paths in 1.95 sec```

File: octoprint_mrbeam/__init__.py (hunk 1 above)
Comment (amrsoll) [reply octoprint_mrbeam/__init__.py]: Then the whole function can be decorated with the `@get_thread(daemon=True)` decorator.

File: octoprint_mrbeam/filemanager/file_storage.py [hunk 6]
```diff
@@ -0,0 +1,89 @@
...
+
+        self._remove_metadata_multiple_entry(metadata_to_remove)
+
+        self._logger_mrb.info(
+            "remove_multiple_files(): Deleted %s/%s files in %s paths in %.2f sec",
+            f_count,
+            len(files),
+            len(metadata_to_remove.keys()),
+            time.time() - ts_start,
+        )
```
amrsoll, line 62: can also use the `@logtime` decorator. More compact but less verbose obviously

amrsoll (approved):
I still think it can be improved using the previous comments, but this works

octoprint_mrbeam/__init__.py
andywerner: Not exactly. Here I make sure that there is only one file deletion thread running at a time.

PR #1102, opened 2020-12-10
Calibration tool should only be accessible if calibration tool mode i…

andywerner:
…s enabled

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -80,7 +80,10 @@ from octoprint_mrbeam.util.material_csv_parser import parse_csv
 from octoprint_mrbeam.util.calibration_marker import CalibrationMarker
 from octoprint_mrbeam.camera.undistort import MIN_MARKER_PIX
 from octoprint_mrbeam.util.device_info import deviceInfo
-from octoprint_mrbeam.util.flask import restricted_unless_calibration_tool_mode
+from octoprint_mrbeam.util.flask import (
+    restricted_unless_calibration_tool_mode,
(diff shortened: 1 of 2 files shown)


amrsoll:
I don't understand what this is for

amrsoll:
I see. The users can find the tool if logged in.

I would modify the existing function though:

```py
def wrap_if(condition=False, decorator=identity):
    """
    Run a decorator function` if ``condition`` is met.
    """
    if condition:
        return decorator
    else:
        return identity

####

def blind_wrap(return_val):
    """ Wrap a function so that it will always return ``return_val`` """
    def wrap(f):
        def puppet(*a, **kw):
            return return_val
        return puppet
    return wrap 

def abort_if(condition=False, code=401):
    return wrap_if(condition, decorator = blind_wrap(abort(code)))

# OR

def abort_if(condition=False, code=401):
    """ I suspect ``abort`` raises an Exception, in which case this works better. """
    def wrap(f):
        def puppet(*a, **kw):
            return abort(code)
        return puppet
    return wrap_if(condition, decorator=wrap)

####

def require_calibration_tool_mode(func):
    @functools.wraps(func)
    def decorated_view(cls_obj, *args, **kwargs):
        return abort_if(not cls_obj.calibration_tool_mode)(func)(
            cls_obj, *args, **kwargs
        )
    return decorated_view
```

andywerner:
I'm find with your suggestion. From what I can see it's just style/aesthetics, right? Feel free to change it your way.

amrsoll:
Screenshot from 2021-01-26 11-10-20

PR #1105, opened 2020-12-14
Add better time estimation to "Prevent unwanted burn marks"

amrsoll:
i don't have a great way to test it right now, but the only problem should be the regex expression, and even then it's simple.

### DIFF
--- a/octoprint_mrbeam/gcodegenerator/jobtimeestimation.py
+++ b/octoprint_mrbeam/gcodegenerator/jobtimeestimation.py
@@ -25,15 +25,27 @@ def time_from_comment(comment):
     Returns the number of seconds read in the comment if the comment matches a float:
         ; [\+-]?[0-9]+\.?[0-9]*s
 
-    Example:
-        ; +1.3s
+    Examples:
+        >>> time_from_comment("0.19s")
+        0.19
+        >>> time_from_comment("; +0.19s")
+        0.19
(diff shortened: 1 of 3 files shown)


octoprint_mrbeam/gcodegenerator/jobtimeestimation.py
```diff
@@ -17,6 +17,35 @@
...
+        >>> time_from_comment("; not formatted 70")
+        0.0
+        >>> time_from_comment("cannot do 2 numbers 6.9s 42.0s")
+        6.9
+    """
+    match = MATCH_COMMENT_ADD_TIME.search(comment)
+    if match:
+        float(match.group(0))
+    else:
+        return 0.0
```
andywerner: Can we have this more explicit with a key word eg: `EXTRA_JOB_TIME:45.2`
This way  it's more clear that there is a system that parses this comment and it prevents humans from adding job time by accident.

I appreciate the documentation though

andywerner (changes requested):

octoprint_mrbeam/gcodegenerator/jobtimeestimation.py
amrsoll: Note to self : It won't parse successfully if the "s" is added -> `group(1)` and change regex

andywerner (approved):

PR #1110, opened 2020-12-29
Click analytics

andywerner:


octoprint_mrbeam/templates/error_logging_script.jinja2
```diff
@@ -29,7 +29,17 @@
 
             let msg = []
             for (let a of arguments) {
-                msg.push((typeof(a) === "object" && a !== null) ? JSON.stringify(a) : a)
+                if (typeof(a) === "object" && a !== null) {
+                    try {
+                        msg.push(JSON.stringify(a))
+                    } catch (e) {
+                        msg.push("<object non-serializable>")
+                        // nothing
```
amrsoll, line 37: Perhaps interesting to add the variable name instead of a blank statement? Or the object keys which are necessarily serialisable?

andywerner: Nice idea. Still, for now I want continue without this improvement.

PR #1117, opened 2021-01-11
Fix image y value when applying matrix transform.

### LINKED ISSUE #1116: Opposite Y value after 2 transforms
Looks like there might be something wrong with the math when transforming the image element in this SVG

In the screenshot, the image has a `y` value of `280` instead of `-280`

Screenshot from 2021-01-11 11-39-51

[Side 1_changed.svg.txt](https://github.com/mrbeam/MrBeamPlugin/files/5795242/Side.1_changed.svg.txt)

amrsoll:
Fixes #1116

### DIFF
--- a/octoprint_mrbeam/static/js/matrix_oven.js
+++ b/octoprint_mrbeam/static/js/matrix_oven.js
@@ -161,6 +161,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
                 elem.attr({
                     style: "transform: scale(1,-1); transform-origin: top",
                     height: -transformedH,
+                    y: -transformedY,
                 });
             }
             return ignoredElements;


octoprint_mrbeam/static/js/matrix_oven.js
```diff
@@ -161,6 +161,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
                 elem.attr({
                     style: "transform: scale(1,-1); transform-origin: top",
                     height: -transformedH,
+                    y: -transformedY,
```
andywerner, line 164: Wondering: if transformedH goes into 'height', shouldn't transformedY go to width?

hungerpirat:
Hey @amrsoll , looks good. Do we have a test for this issue now?

amrsoll:
I think we should merge the testing branch before I change a test to show the fix : #1026 

There is already a test for that file there, but I should change the name to reflect the nature of the change

andywerner (approved):

octoprint_mrbeam/static/js/matrix_oven.js
amrsoll: I do not understand why it would? `transformedH` stands for 'transformed Height', and we only modify the Y axis with the `transform: scale(1, -1)`, no need to touch the width

PR #1125, opened 2021-01-22
fixes #1124: placeSVG: unref creates duplicate ids and removes unreferenced elements

### LINKED ISSUE #1124: placeSVG: unref creates duplicate ids and removes unreferenced elements
When placing an SVG ono the working area, unref() is called to replace all use-elements by it's referenced element.

If the referenced element has an id attribute and is referenced more than once, unref() will produce duplicate ids and the removes all but the first unreferenced element.

Reproduce with this SVG:
[carina neu.svg.zip](https://github.com/mrbeam/MrBeamPlugin/files/5857485/carina.neu.svg.zip)

andywerner:
fixes #1124

### DIFF
--- a/octoprint_mrbeam/static/js/unref.js
+++ b/octoprint_mrbeam/static/js/unref.js
@@ -38,7 +38,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
         // 2. replace them and remember the referenced source elements
         for (var i = 0; i < elements_to_replace.length; i++) {
             var e = elements_to_replace[i];
-            var src = e._replace_with_src();
+            var src = e._replace_with_src(i);
             // src might be undefined in some error cases.
             // I intentionally let it crash here to "fail fast" since the user gets an error message.
             used_source_elements.push(src);
@@ -65,7 +65,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
      *
      * @returns {url} : url of the used src element
      */
-    Element.prototype._replace_with_src = function () {
+    Element.prototype._replace_with_src = function (counter) {
         var elem = this;
         if (elem.type !== "use") {
             return;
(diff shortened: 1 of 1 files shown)


hungerpirat:
👏👏👏👏

octoprint_mrbeam/static/js/unref.js
```diff
@@ -138,7 +144,9 @@ Snap.plugin(function (Snap, Element, Paper, global) {
             ];
             for (var i = 0; i < attribute_keys.length; i++) {
                 var key = attribute_keys[i];
-                if (attr[key]) new_attrs[key] = attr[key];
+                if (elem_attr[key]) {
+                    src_attrs[key] = elem_attr[key];
+                }
```
amrsoll, line 149: Shouldn't it be the other way around? 

`elem_attr[key] = src_attrs[key]`

Otherwise you would be modifying the source element instead of the newly created element.

amrsoll: if you do not want the `src_attr` to replace `elem_attr` keys then you should prune `src_attrs` instead, and that can be done without using a long list of keys.

amrsoll (changes requested):
Admittedly the changes are fixing the issue, but I see that there can be a potential bug in that function which should be fixed.

octoprint_mrbeam/static/js/unref.js
amrsoll: I am not quite sure if the `src_attr` is a copy or just a pointer, which might influence the decision to change the function.

amrsoll: @hungerpirat ?

andywerner: src_attr is a copy
Attributes of the placeholder element (here elem) should override attributes of the library element (here src_elem or src_attrs).Then the placeholder is replaced with a copy of the library element, this copy has all attrs of the original placeholder.

amrsoll: yes but you don't need to keep a giant copy of all possible overlapping keys. Either replace the relevant keys (the keys they have in common except `id`) or remove the keys they have in common (and `id`) so that they don't overlap when updating the keys. 
I presume `uplicate.attr(src_attrs)` updates the keys? Or does it overwrite them?

amrsoll: If you decide to keep the list, then you should remove `id` because it might overwrite your previous work on `src_attrs["id"]`

```js
            if (src_attrs["id"] !== undefined) {
                src_attrs["id"] = src_attrs["id"] + "_unref_" + counter;
            }
```

hungerpirat (approved):
Looks good for me. I did not test it - I am sure @Gallore did. Approved.

octoprint_mrbeam/static/js/unref.js
andywerner: Everything as planned. If the placeholder already has an ID, we keep this one.

andywerner: After all I did not want to refactor this cod here. I just wanted to fix a bug in the existing design.

Issue #1137, opened 2021-02-02
NPS: Add more data to user review [/1]


Gallore:
When we submit a user review ("Would you recommend Mr Beam to your friends? 1-10?") we need to add more device data and also ask the user more details.

Device data:
- SW version (mr beam plugin)
- model (MRBEAM2_DC, ...)
- production date (be aware that older devices do not have a model_id or production date)

Server side:
We need to add the date when the review was uploaded (which might be very different from the date when the review was done.) The creation date of the json file in the GCP bucket could be used. All following components need to user or forward this date.

User data:
- We should ask the user if it's ok to contact her/him "in case we have questions" and offer to add email address or phone number. Could be one (free text) field or two fields. We could also pre-fill the user's email address, but I lean towards not doing this. We probably also need a statement what we use these data for. 
- the user should have the option to change it's rating. Since we asks in any case for feedback and contact details, it does not change the logic/flow of the user review screen.
- please remove the "wow thank you" text as if we were surprised about a good rating. Let's have a neutral "Thank you for your feedback" or "Your feedback is super important to us" or so.
- (Optional) If the user gave us a good rating, we could add another screen AFTER he submitted his complete feedback. This new screen would be dedicated to suggest that the suer rates us on rating pages. this page should get a visual makeover I think. In my mind we should show there logos of well known rating platforms. But I need to check with Jessie. I see this screen as optional, we can add it later on as well.
  - Our Shop (as it is currently, but please let's use a redirect `mr-beam.org/beamos/review`) 
  - Google
  - Trustpilot

Gallore:
Device data and server side fixed in commit 915d4d8

Gallore:
@khaledsherkawi please focus un the part "User data" in the description. The other things are already done.

Gallore:
@khaledsherkawi Please KEEP the switch for users who rated 7 and better!
We need a switch AFTER the suer entered his feedback and submitted it:
- <7 "Thank you for your feedback"
- >= 7: "Thank you for your feedback" + Please give us a review. For now this screen could be simple - just what we have now. Bonus points if the call to action (the review link) is a bit more visible. Later on (probably not for this release) we redesignt this scree according to how I described it in the description.

Gallore:
Wording:

 **Screen 1**

Title: 
`How likely are you to recommend Mr Beam to a friend or colleague?`
[...]

PR #1141, opened 2021-02-03
Adds more device data to a user review (#1137) [/2]

andywerner:


octoprint_mrbeam/static/js/review.js
```diff
@@ -147,16 +153,19 @@ $(function () {
         self.sendReviewToServer = function () {
             let review = $("#review_textarea").val();
             let data = {
+                // more data is added by the backend
                 dontShowAgain: self.dontShowAgain(),
                 rating: self.rating(),
                 review: review,
                 ts: new Date().getTime(),
-                env: MRBEAM_ENV_LOCAL,
                 sw_tier: MRBEAM_SW_TIER,
```
amrsoll: Not necessary anymore

octoprint_mrbeam/static/js/review.js
```diff
@@ -147,16 +153,19 @@ $(function () {
...
-                snr: MRBEAM_SERIAL,
                 number: self.REVIEW_NUMBER,
             };
 
+            if (FORCE_REVIEW_SCREEN) {
+                data["debug"] = true;
+            }
```
amrsoll, line 166: Why not this :
```js
data ={
  ...
  debug: FORCE_REVIEW_NUMBER
)
```

octoprint_mrbeam/analytics/review_handler.py
```diff
@@ -48,11 +49,28 @@ def _on_mrbeam_plugin_initialized(self, event, payload):
...
+        given = self._settings.get(["review", "given"])
+        if not given or data.get("debug", False):
+            data = self._add_review_data(data)
+            self._write_review_to_file(data)
+            self._settings.set_boolean(["review", "given"], data["dontShowAgain"])
+            self._settings.save()  # This is necessary because without it the value is not saved
+
+            ReviewFileUploader.upload_now(self._plugin, self._review_lock)
+        else:
+            self._logger.warn("Not accepting user review since it was already given.")
```
amrsoll, line 61: How about updating the dataset with `data["already_given"] = given` and filter afterwards, perhaps to see if the review changed

amrsoll (approved):

octoprint_mrbeam/static/js/review.js
andywerner: true

octoprint_mrbeam/analytics/review_handler.py
andywerner: Each device should only send one review ever (only exception: if the device got reset). So far this was only enforced on the frontend side while the backend accepted happily as many reviews as the frontend (or a dev tools hacker) submitted. This is just to enforce this policy on the backend as well. Therefor I thinks it's ok this way.

octoprint_mrbeam/static/js/review.js
andywerner: se my comment above

PR #1145, opened 2021-02-04
Air filter run-on phase 2 is now on 15% (before auto)

### LINKED ISSUE #1142: Set Air Filter run-on (after a laser job) to 10%
If the dust sensor is dirty, fan auto mode is still very strong/loud. In this case the "cancel air filter run on" button at the end of a laser job does seem to make any sense to the user, since the difference between 100% and auto mode is not noticeable. As a pragmatic fix, Teja suggested to set it run on to 10% instead of auto mode.

andywerner:
fixes #1142


amrsoll (approved):
Shame that the auto mode doesn't work

andywerner:
> Shame that the auto mode doesn't work

yes... :-(

PR #1150, opened 2021-02-10
v0.8.0.2: Analytics logrotate allows files up to 100M now

andywerner:
Since we addec lick analytics I think this is important to no loos too much analytics data


amrsoll (commented):
Doesn't it risk creating errors if you run the logrotate migration fix a second time on the devices?

Reason : There is an "mv" command that will create an error because it won't find the file that it moved the first time.

andywerner:
> Doesn't it risk creating errors if you run the logrotate migration fix a second time on the devices?
> 
> Reason : There is an "mv" command that will create an error because it won't find the file that it moved the first time.

Good point. But nope. Just re-tested and the system does not complain. All good

PR #1151, opened 2021-02-10
Addition of user details to review form

### LINKED ISSUE #1137: NPS: Add more data to user review [/1]

khaledsherkawi:


andywerner (changes requested):
Please change as we discussed today on the phone.
In the issue I wrote the wording : https://github.com/mrbeam/MrBeamPlugin/issues/1137#issuecomment-776919717

khaledsherkawi:
The only thing I kept from before is the back button between screens 1 and 2. The rating bar doesn't look good on top of screen 2. In addition the CSS of the rating bar (on top of screen 2) will be confusing to the user. If we leave it without the their decision highlighted they will think that they have to vote again and if we highlight their decision they will think they cannot change it. With the back button they can choose to go back and answer differently. 

Screen 1:
Screen 1

Screen 2:
Screen 2

Screen 3 - positive review:
Screen 3 - Positive Review

Screen 3 - negative review:
Screen 3 - Negative Review

octoprint_mrbeam/templates/review.jinja2
```diff
@@ -5,6 +5,7 @@
             <h3 id="review_question" class="review_title">{{ _('How likely are you to recommend Mr Beam to a friend or colleague?') }}</h3>
             <div id="rating_block">
                 <div class="rating">
+                    <span class="rating-label">Not at all likely</span>
```
andywerner: pls mark for translation

octoprint_mrbeam/templates/review.jinja2
```diff
@@ -15,23 +16,47 @@
                     <button value="8">8</button>
                     <button value="9">9</button>
                     <button value="10">10</button>
+                    <span class="rating-label">Extremely likely</span>
```
andywerner: pls mark for translation

octoprint_mrbeam/templates/review.jinja2
```diff
@@ -15,23 +16,47 @@
...
+            <div id="ask_user_details" class="review_hidden_part">
+                <label for="review_input_phone_email">{{ _('Phone or email: (optional)') }}
+                    <small>{{ _('If it\'s ok to contact you about your feedback.') }}</small>
+                </label>
+                <input type="text" class="form-control" id="review_input_phone_email">
+            </div>
+            <div id="positive_review" class="review_hidden_part">
+                <h3 class="review_title">{{ _('Thank you for your response.') }}</h3>
+                <p>{{ _('Positive reviews from awesome customers like you help others to feel confident about choosing Mr Beam too.') }}</p>
+                <p>{{ _('Could you take 60 seconds to go to the ') }}
```
andywerner: Please bring this whole sentence in one translatable entity with placeholders for opening and closing tag. This is This is important for translations.

octoprint_mrbeam/templates/review.jinja2
```diff
@@ -15,23 +16,47 @@
...
+                    <a id="give_review_link" target="_blank" href="https://store.mr-beam.org/products/mr-beam-ii-dreamcut-air-filter-ii-bundle?utm_source=beamos&utm_medium=beamos&utm_campaign=review_screen#shopify-product-reviews">{{ _('Mr Beam Store') }}</a>
```
andywerner: please use this URL: `https://mr-beam.org/beamos/review`

andywerner (changes requested):
Here is a description on how to mark text in jinja files: https://sites.google.com/mr-beam.org/wiki/development/internationalization?authuser=0#h.p_lMBteIJOQst1

andywerner (approved):

PR #1152, opened 2021-02-10
Messaging System

khaledsherkawi:


khaledsherkawi:
new notification
no messages selected
message template no images
message template with images

PR #1154, opened 2021-02-11
Fix the NaN error for an overshoot + tests

### LINKED ISSUE #1134: Extra Overshoot creates NaN coordinates [2/]
SP : 2
In some cases the gcode produces for an overshoot creates NaN coordinates such as `G0XnanYnan`.

It only happens when an extra overshoot manoeuvre is requested and it might occur when the overshoot distance is set to 0, as it expects the size of the overshoot to scale with that value, but it also a modulo ` something % size *... ` which creates a `np.nan` in the numpy library.

amrsoll:
Fixes #1134 

```
# TODO integrate octogon overshoots in write_gcode_for_line_start
```
@hungerpirat 
I purposefully did Not put the overshoot logic inside any function that deals with lasering a line in the image, because it Only happens on a transition from line to line. I am further going to say that the classic overshoot system with the line_start and line_end is not functional because it adds extra gcode in place they don't need to be and it splits the whole overshoot comprehension in different places of the code, essentially obfuscating it.


amrsoll:
All the tests are now passing

PR #1158, opened 2021-02-12
German translation

andywerner:


andywerner:
This commit message is incorrect:
> Removed empty stings from translation (caused issues), corrected typo… …
cb3f156s, German translation

Should be:
> German translation for additional material hints

khaledsherkawi (approved):

PR #1168, opened 2021-02-18
F rasterization splitup

hungerpirat:
All to raster elements are rendered separately.
+ less data transmission between browser & beam
+ smaller images to process in the backend
+ faster engravings as designs multiplied to a grid are now processed one by another instead of all across the working area
+ slicker rendering pipeline using promises & async await
+ white background addition now in the img2gcode.py again: better UX when using .mrb files, faster due to smaller images.
+ code cleanup. snap_helpers.js introduced for general snap plugins.
+ bugfixes:
- matrix_oven: bake without callback possible again
- circular dependency in JSON generation in convert.js
- proper calculation of transformed bounding box 
- img2gcode.py rasterengraving start is now lower left of an image instead of upper left


octoprint_mrbeam/analytics/cpu.py
```diff
@@ -142,8 +142,8 @@ def get_cpu_throttle_warnings(self):
             return res
 
         except:
-            self._logger.exception(
-                "Exception in _get_cpu_throttle_warnings while converting cpu get_throttled: %s",
+            self._logger.error(
+                "Exception in _get_cpu_throttle_warnings while converting cpu get_throttled: '%s'",
                 t_hex,
             )
```
amrsoll, line 148: Please log the Exception itself as well

octoprint_mrbeam/gcodegenerator/img2gcode.py
```diff
@@ -252,48 +252,62 @@ def img_prepare(self, img, w_mm, h_mm):
...
+        bbox = img.getbbox()
+        #        self.log.info("#####")
+        #        self.log.info(bbox)
+        #        self.log.info((0, 0, dest_wpx, dest_hpx))
+
+        # 1a. crop to bbox
+        if True:
+            self.profiler.start("crop")
+
+            # TODO: this removes only transparent pixels, white pixels are still counted as content.
```
amrsoll, line 266: Perhaps we should convert the white to alpha first then to minimise processing time
https://stackoverflow.com/a/54148416/11136955

amrsoll: Or if you invert the image colors, the white background becomes black (aka zeroes) and I presume the [getbbox](https://pillow.readthedocs.io/en/4.1.x/reference/Image.html#PIL.Image.Image.getbbox) method would filter them out?
> Calculates the bounding box of the non-zero regions in the image.

octoprint_mrbeam/gcodegenerator/img2gcode.py
```diff
@@ -252,48 +252,62 @@ def img_prepare(self, img, w_mm, h_mm):
...
+                )
+                return []
+
+            left, upper, right, lower = bbox  # bbox is a tuple of four
+            bb_w = right - left
+            bb_h = lower - upper
+            if bb_w != dest_wpx or bb_h != dest_hpx:
+                img = img.crop(bbox)
+                old_pixels = dest_wpx * dest_hpx
+                bb_area = bb_w * bb_h
```
amrsoll, line 280: please keep name consistency like `old_area` / `new_area`

octoprint_mrbeam/gcodegenerator/img2gcode.py
```diff
@@ -252,48 +252,62 @@ def img_prepare(self, img, w_mm, h_mm):
...
-        # 					img.save("/tmp/img2gcode_2_whitebg.png")
-        #
-        # 			self.profiler.stop('remove_transparency').start('contrast')
+        bbox = img.getbbox()
+        #        self.log.info("#####")
+        #        self.log.info(bbox)
+        #        self.log.info((0, 0, dest_wpx, dest_hpx))
+
+        # 1a. crop to bbox
+        if True:
```
amrsoll, line 263: what for?

octoprint_mrbeam/gcodegenerator/img2gcode.py
```diff
@@ -252,48 +252,62 @@ def img_prepare(self, img, w_mm, h_mm):
...
+
+            self.profiler.stop("crop").start("remove_transparency")
+
+        # 2. remove transparency
+        if (not self.is_inverted) and (img.mode == "RGBA"):
+            whitebg = Image.new("RGBA", (dest_wpx, dest_hpx), "white")
+            img = Image.alpha_composite(whitebg, img)
+
+            if self.debugPreprocessing:
+                img.save("/tmp/img2gcode_2_whitebg.png")
```
amrsoll, line 310: Could it be that - if we do this before cropping - it might be easier to get the bbox (with inverted colors so the background is black) ?

octoprint_mrbeam/gcodegenerator/img2gcode.py
```diff
@@ -477,8 +491,16 @@ def generate_gcode(self, imgArray, xMM, yMM, wMM, hMM, file_id):
...
+            x_start = x_off
+            y_start = (
+                y_off - height_px * self.beam
+            )  # lower left of the image, but with safety whitespace around the content
             gc = self._get_gcode_g0(
-                x=x_off, y=y_off, comment="; Move to start ({},{})".format(x_off, y_off)
+                x=x_start,
+                y=y_start,
+                comment="; Move to start ({},{})".format(x_start, y_start),
             )
```
amrsoll: Well perhaps we don't need this if the line-by-line image 2 gcode function already give a `G0` gcode command to move to the first colored pixel (with a speed ramp distance or backlash compensation)

octoprint_mrbeam/static/js/helpers/working_area_helper.js
```diff
@@ -308,29 +309,49 @@ class WorkingAreaHelper {
...
         return true;
     }
+
+    static _getSubstringAtIndex(input, index, delimiter = "[^a-zA-Z0-9.,]+") {
+        const iter = input.matchAll(delimiter);
+        let idx1 = 0;
+        let idx2 = input.length;
+        let del;
+        while ((del = iter.next()) !== null) {
+            if (del.done) break;
```
amrsoll, line 345: refactoring suggestion
```js
let del = iter.next();
while ( del !== null && ! del.done) {
 ...
  del = iter.next();
}
```

octoprint_mrbeam/static/js/helpers/working_area_helper.js
```diff
@@ -308,29 +309,49 @@ class WorkingAreaHelper {
...
+                idx2 = del.value.index;
+                break;
+            }
+        }
+        const strBefore = input.substring(0, idx1);
+        const v1 = input.substring(idx1, idx2);
+        const strAfter = input.substring(idx2);
+        //console.debug("idx, idx1 -> idx2", index, idx1, idx2);
+        return [strBefore, v1, strAfter];
+    }
```
amrsoll, line 358: That was a mouthfull :open_mouth: 

I would recommend doing something like so (pseudo Js code)

```js
before_cursor = input.substr(0, index).split(delimiter)
after_cursor = input.substr(index).split(delimiter)
focused_value = float(before_cursor.pop(-1) + after_cursor.pop())
new_string = chain(before_cursor, str(focused_value + shift), after_cursor).join(default_delimiter)
```

octoprint_mrbeam/static/js/render_fills.js
```diff
@@ -53,18 +98,84 @@ Snap.plugin(function (Snap, Element, Paper, global) {
...
             ) {
-                selection.push(elem);
+                if (elem.type === "#text") {
+                    let parent = elem.parent();
+                    parent.addClass(className);
+                    selection.push(parent);
+                } else {
+                    elem.addClass(className);
+                    selection.push(elem);
+                }
```
amrsoll, line 108: Please simplify

```js
let _elem;
if (elem.type === "#text")
  _elem = elem.parent();
else
  _elem = elem
_elem.addClass(className)
selection.push(_elem)
```
Or is there an immutability issue I'm not seeing?

octoprint_mrbeam/static/js/render_fills.js
```diff
@@ -53,18 +98,84 @@ Snap.plugin(function (Snap, Element, Paper, global) {
...
             } else {
                 if (fillPaths && elem.is_filled()) {
+                    elem.addClass(className);
                     selection.push(elem);
```
amrsoll, line 112: This whole if statement chain should probably be:

```js
if (elem.type === "#text") {
  let parent ...
} else if (elem.type === "text" || elem.type === "image" || (fillPaths && elem.is_filled()) {
 ... addClass and push into selection
} else {
  selection.remove(elem)
}

File: octoprint_mrbeam/static/js/render_fills.js [hunk 11]
```diff
@@ -53,18 +98,84 @@ Snap.plugin(function (Snap, Element, Paper, global) {
...
+        let clusterCount = 0;
+        let clusters = [];
+        for (let i = 0; i < marked.length; i++) {
+            let rasterEl = marked[i];
+            const bbox = rasterEl.get_total_bbox();
+            // find overlaps
+            let lastOverlap = -1;
+            for (var j = 0; j < clusters.length; j++) {
+                var cluster = clusters[j];
+                if (Snap.path.isBBoxIntersect(cluster.bbox, bbox)) {
```
amrsoll, line 133: Couldn't it be that distinct elements can still be identified as overlapping if their bounding boxes collide?

I am thinking for example of 2 parallel lines going in a diagonal

If they are rasters at this stage - The way I solved this in openCV was to check if there was no common pixel between 2 clusters (by ordering the lists of pixel indexes and comparings the 2 ordered lists, or you can concatenate the 2 lists of indexes and find duplicates which is essentially the same algorithm)

amrsoll: or perhaps `svg` has already something for determining connected / overlapping elements?

octoprint_mrbeam/static/js/render_fills.js
```diff
@@ -247,89 +359,121 @@ Snap.plugin(function (Snap, Element, Paper, global) {
...
+        const root = this;
+        let elemsToFix = root.selectAll(selector);
+        for (let i = 0; i < elemsToFix.length; i++) {
+            const e = elemsToFix[i];
+            const originalId = e.attr(srcIdAttr);
+            if (originalId !== null && originalId !== "") {
+                e.attr({ id: originalId });
+            }
+            //console.log(`fixed Id: ${e.type}#${originalId}`);
+        }
```
amrsoll, line 476: I would recommend using map here

```js
elemsToFix = elemsToFix.map(e => e.attr({id: e.attr(srcIdAttr)})
```
I don't know if a `null` id will mess things up, otherwise it should be possible to make an if statement in a separate functions.
```js
function fixId(elm) {
  if (elm.attr(srcIdAttr))
    elm.attr({id: elm.attr(srcIdAttr)});
}
let elemsToFix = root.selectAll(selector).map(fixId)
```

octoprint_mrbeam/static/js/working_area.js
```diff
@@ -2490,80 +2503,101 @@ $(function () {
...
+            // embed filters
+            // copy defs for filters (e.g. imgCropping, imgSharpening, etc...)
+            var originalFilters = snap.selectAll("defs>filter");
+            for (let i = 0; i < originalFilters.length; i++) {
+                const original_id = originalFilters[i].attr("id");
+                const clone = originalFilters[i].clone();
+                const destFilter = clone.appendTo(targetDefs);
+                // restore id to keep references working
+                destFilter.attr({ id: original_id, "mb:id": original_id });
+            }
```
amrsoll, line 2558: what is this for? It only seems to modify a clone and then the clone is not used

amrsoll (changes requested):
Very interesting, I really like the changes!

octoprint_mrbeam/static/js/helpers/working_area_helper.js
amrsoll: That also changes any weird separator to the default separator

amrsoll: Found a nice reduce one-liner for `new_string`
```js
let default_delimiter = ','
[1, 2, 3].concat(5, [6, 7]).reduce((x, y) => x + default_delimiter + y)
// "1,2,3,5,6,7"
```

octoprint_mrbeam/analytics/cpu.py
hungerpirat: Is it really valuable information? What is it useful for? 
I decided to not log the exception as it was spamming my logfile with stacktraces without any helpful information. 
IMHO the stacktrace is not necessary here, as the code is not wrong - just the sensor delivered bad data. Exactly the reason why we are catching exceptions here.

amrsoll: Ah I see ok

octoprint_mrbeam/static/js/helpers/working_area_helper.js
amrsoll: One should probably also filter out the empty values after the split (in case of 2 adjacent separators) :
```js
before_split = before_cursor.split(delimiter).filter(e=>e)
```

amrsoll: The main issue with replacing the separator is that the cursor might jump from one value to the next after incrementing

octoprint_mrbeam/gcodegenerator/img2gcode.py
hungerpirat: avoiding stupid laserhead moves when starting an image engraving.

octoprint_mrbeam/gcodegenerator/img2gcode.py
hungerpirat: processing time of this step is 0.5% of the total conversion time. Not worth to optimize.

octoprint_mrbeam/gcodegenerator/img2gcode.py
hungerpirat: go for it!

octoprint_mrbeam/gcodegenerator/img2gcode.py
hungerpirat: i don't see any need for an optimization here.

octoprint_mrbeam/gcodegenerator/img2gcode.py
hungerpirat: We do need this to ramp up the laser driver. 
Basically this is a bugfix. The change is only in the coordinates - and yes, it still could be improved by figuring out the real first pixel of the image. But I skipped further optimization as our resources are sparse and there is no business benefit. Additionally code stays more modular.

octoprint_mrbeam/static/js/helpers/working_area_helper.js
hungerpirat: I don't care which style a loop is written. Can't see any serious advantage on one of those solutions.

octoprint_mrbeam/static/js/helpers/working_area_helper.js
hungerpirat: What's the benefit ?

hungerpirat: > That also changes any weird separator to the default separator

? what is a weird separator ?

hungerpirat: > Found a nice reduce one-liner for `new_string`
> 
> ```js
> let default_delimiter = ','
> [1, 2, 3].concat(5, [6, 7]).reduce((x, y) => x + default_delimiter + y)
> // "1,2,3,5,6,7"
> ```

does this really do what _getSubstringAtIndex() is doing?  I don't understand...

octoprint_mrbeam/static/js/render_fills.js
hungerpirat: both is fine for me. my one is tested pretty often, yours still needs to be tested.

octoprint_mrbeam/static/js/render_fills.js
hungerpirat: is that really simpler to understand?

octoprint_mrbeam/static/js/render_fills.js
hungerpirat: you're right. cluster overlapping by checking bounding boxes is not giving a precise result. 
It's only a 90% solution. 
To solve the other 10% you have to deal with overlaps of all combination of paths, images, text, cropped images, ... 
Snap doesn't have proper tools for that. 
This 90% solution is giving a 500% boost for the users who are filling up their working area for mass production. The final 10% ... What is the dev/testing/maintenance effort? What is the extra benefit for the user?

octoprint_mrbeam/static/js/render_fills.js
hungerpirat: And what is the benefit if we rewrite and retest it this way? 
For us? For the user? At what cost?

octoprint_mrbeam/static/js/working_area.js
hungerpirat: You need to learn more about Snapsvg. A clone is directly hooked in the DOM. No need to "use" it.

octoprint_mrbeam/gcodegenerator/img2gcode.py
amrsoll: No I meant the `if True`

octoprint_mrbeam/gcodegenerator/img2gcode.py
amrsoll: Yeah but I mean, when we do the image scan it's already creating a move command (including overshoot distance) before lasering the first pixels.

octoprint_mrbeam/static/js/helpers/working_area_helper.js
amrsoll: The benefit is code readability, super important!

By weird separator I mean anything that isn't what we expect like a `////` as the default `delimitor` variable would suggest.

It's not meant to just replace this specific function but also part of the code where the function is referenced.

What it does is
- parsing the line with the given separators
- selecting the item you have your cursor on
- increment it with some value
- join together the previously parsed items with a default separator (can be different than what the user typed in)

octoprint_mrbeam/gcodegenerator/img2gcode.py
amrsoll: It might not even be starting on the correct line in the case of clusters.

octoprint_mrbeam/static/js/helpers/working_area_helper.js
amrsoll: I see that it's kind of a thing in javascript apparently
Whenever I see the return value of a variable assignment being used, I start seeing red flags because of how it is missued and dangerous in other older languages.

octoprint_mrbeam/static/js/render_fills.js
amrsoll: It's all about readability. The code is read way more often than it is being written/ rewritten, therefore readability makes it easier on ourselves.

N.B. also somewhat performance because maps can be processed faster, but that's not the issue here.

octoprint_mrbeam/static/js/render_fills.js
amrsoll: yeah debatable :)

octoprint_mrbeam/static/js/render_fills.js
amrsoll: Alright I was just being curious

andywerner:
fixes #1229

PR #1171, opened 2021-02-19
Removes login credentials from frontend.log logfile

andywerner:
Axel:
:warnung: We are tracking sensitive information now - like plaintext username / password combos through the front end and the log of the front end. :warnung:
The context that I have found is an analysis zip from a user. It would seem they had put in the wrong password and it shows in the logfile...


amrsoll (approved):
Seems good. I'm wondering if the password could be hashed in the frontend before sending to the device :thinking:

PR #1173, opened 2021-02-22
Fix user permissions for calibration and clean left over files.

### LINKED ISSUE #1146: Purge the detection result for all of the images taken during the lens calibration
The .npz files for each chessboards remain after exiting the lens calibration. This breaks the calibration when restarting it from scratch

### LINKED ISSUE #1165: Remove results when restarting a Lens Calibraton [2]
When restarting a Lens Calibration, the .npz files for every chessboard picture taken are not properly deleted. This means that it affects the following calibration as it will take into account the files from the failed calibration.

### LINKED ISSUE #1167: Fix permissions for the Lens Calibration
When starting the Lens calibration, the permission is not granted Err 401

TODO
- [ ] Allow the proper functions to be called when not in factory mode
- [ ] Document which functions get called for the calibration
- [ ] Add testing to make sure that the lens calibration is at least available when logged in (Selenium ?) and that the factory page is not accessible when not in that mode (Should return error 401).

amrsoll:
Fixes #1146 
Fixes #1165 
Fixes #1167

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -83,7 +83,10 @@ from octoprint_mrbeam.util.cmd_exec import exec_cmd, exec_cmd_output
 from octoprint_mrbeam.util.device_info import deviceInfo
 from octoprint_mrbeam.util.log import logExceptions
 from octoprint_mrbeam.util.material_csv_parser import parse_csv
-from octoprint_mrbeam.util.flask import calibration_tool_mode_only
+from octoprint_mrbeam.util.flask import (
+    calibration_tool_mode_only,
+    restricted_access_or_calibration_tool_mode,
+)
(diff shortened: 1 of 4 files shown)


octoprint_mrbeam/__init__.py
```diff
@@ -1345,33 +1344,28 @@ def takeUndistortedPictureForInitialCalibration(self):
     @octoprint.plugin.BlueprintPlugin.route(
         "/on_camera_picture_transfer", methods=["GET"]
     )
-    @calibration_tool_mode_only
```
andywerner, line 1348: now it's open to the world. should be: restricted access OR calibration tool mode

andywerner (changes requested):

andywerner (approved):

PR #1179, opened 2021-02-24
B material height too large

hungerpirat:
fb user request. after accidentally entering 480 mm custom material height, the ui was broken. 
this bugfix limits the entered height to a maximum of 38mm. additionally it limits the displayed height and adds an own .overThick style if the custom material is above 20mm to save display space.

### DIFF
--- a/octoprint_mrbeam/static/css/tab_workingarea.css
+++ b/octoprint_mrbeam/static/css/tab_workingarea.css
@@ -91,6 +91,7 @@ svg g#crosshair {
 
 svg g#camera_markers circle {
   opacity: 0;
+  pointer-events: none;
 }
 
 circle#marker_preview_pane,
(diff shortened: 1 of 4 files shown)


octoprint_mrbeam/static/css/tab_workingarea.css
```diff
@@ -759,4 +762,5 @@ body.ctrlKey #translateHandle {
   stroke-dasharray: none;
   stroke-dashoffset: 0;
   stroke-opacity: 1;
+  pointer-events: none;
```
amrsoll, line 765: This is the fix for #1178 (same commit) - This is not an issue usually with classical merges, but can be problematic with our squash merge practice.

amrsoll (commented):
I don't really understand. There doesn't seem to be any limitation on the Object height input? It seems that you are only touching on the material thickness which is different.

octoprint_mrbeam/static/css/tab_workingarea.css
andywerner: agree

andywerner (approved):

PR #1185, opened 2021-02-26
Better logging on hw malfunctions

andywerner:


octoprint_mrbeam/iobeam/hw_malfunction_handler.py
```diff
@@ -118,3 +118,6 @@ def show_hw_malfunction_notification(self):
             )
 
         self._user_notification_system.show_notifications(notifications)
+
+    def get_messages_to_show(self):
+        return self._messages_to_show
```
amrsoll, line 123: Suggestion:
The pythonic way of doing this is
```py
@property
def messages_to_show(self):
    return self._messages_to_show
```
And you can use it as a public attribute rather than a class function (meant to make it easy to switch a public attribute to private)
```python
something = self.messages_to_show
```

amrsoll: N.B to create the set function :
```python
@messages_to_show.setter
def messages_to_show(self, value):
    self._messages_to_show = value

# And can be used as a simple attribute assignment
self.messages_to_show = new_value
```

amrsoll (approved):
see comments :)

PR #1206, opened 2021-03-17
Object height more visible

andywerner:


amrsoll:
At this point, I feel it would be best to take it outside the Preview tab totally. Now we are adding a lot of text.
Perhaps we should have a specific icon and an info box that triggers when you hover over it?

PR #1216, opened 2021-03-25
fixes issue with unicode in mrb_logger that prevented slicing.

andywerner:
Also added try/catch in do_slice()

### DIFF
--- a/octoprint_mrbeam/mrb_logger.py
+++ b/octoprint_mrbeam/mrb_logger.py
@@ -82,7 +82,12 @@ class MrbLogger(object):
         :param terminal_dump: Collect and log a terminal dump. Terminal dumps are also sent to analytics if analytics is not explicitly set to False.
         :type kwargs:
         """
-        msg = unicode(msg, "utf-8")
+
+        try:
+            msg = unicode(msg, "utf-8")
+        except TypeError:
+            # If it's already unicode we get this TypeError
+            pass
         if kwargs.pop("terminal", True if level >= logging.WARN else False):
             self._terminal(level, msg, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


octoprint_mrbeam/mrb_logger.py
```diff
@@ -82,7 +82,12 @@ def log(self, level, msg, *args, **kwargs):
         :param terminal_dump: Collect and log a terminal dump. Terminal dumps are also sent to analytics if analytics is not explicitly set to False.
         :type kwargs:
         """
-        msg = unicode(msg, "utf-8")
+
+        try:
+            msg = unicode(msg, "utf-8")
+        except TypeError:
+            # If it's already unicode we get this TypeError
```
andywerner, line 89: @amrsoll Axel i want you also to know this!

amrsoll: I would go for
```py
if not isinstance(msg, unicode):
  msg = unicode(msg, "utf-8")
```
simpler, no comment needed

amrsoll (commented):
I really think we should be using `@logExceptions` instead... If it is an unforseen Exception, wee need to see the stacktrace anyway, and the stacktrace analysis can be done on the analytics server instead.

octoprint_mrbeam/mrb_logger.py
andywerner: @amrsoll:
Turns out parts of the code do pass other things than strings here (lists) and it's fine for the logger. Therefore we must keep it with the exception.
In a general level I'm with you on this philosophy to test if an action is possible before doing the action over just doing it and catching possible exceptions...

PR #1220, opened 2021-03-27
Engraving passes + progressive options added

khaledsherkawi:


octoprint_mrbeam/templates/conversion_dialog.jinja2
```diff
@@ -287,6 +287,29 @@
...
+                                        <span style="display: inline-block">
+                                            <label data-bind="visible: showExpertSettings">
+                                                <input data-bind="checked: engravingProgressive, event: {change: flag_customized_material}" type="checkbox" class="param_eng_progressive"/>
+                                                {{ _('progressive') }}
+                                            </label>
+                                        </span>
+                                    </div>
+                                </div>
+                            </div>
+
```
amrsoll: I think multi-pass engraving would be best suited as a part of the advanced settings as well.

- Change the text to say "engrave" instead of "cut"
- The progressive engraving is probably a bad idea, because it makes it difficult to foresee how deep the engraving gets (progressive cut makes 1 quick pass, then increasingly slow passes), and I don't think we get the same benefits as we would when cutting.

amrsoll (changes requested):
I see `.passes_input` css class isn't used that much, perhaps we could just delete/replace references to it :)

khaledsherkawi:
This is finalized. @amrsoll @Gallore  Please review.

andywerner (approved):

PR #1221, opened 2021-03-29
Add new iobeam logrotate file and refactor migrate.py.

amrsoll:
This adds a new logrotate file that tracks the `.octoprint/log/iobeam.log`, but does not remove the previous one, so iobeam can log in both locations.

Also the refactoring of `migrate.py` improves readability and makes it more compact, especially if black becomes more leniant (need to change black version in pre-commit)


octoprint_mrbeam/files/migrate/iobeam_local.logrotate
```diff
@@ -0,0 +1,9 @@
+/home/pi/.octoprint/logs/iobeam.log {
```
andywerner, line 1: please use same logic:
```
/var/log/iobeam.log {
    size 2M
    rotate 4
    compress
    delaycompress
    missingok
    notifempty
    copytruncate
}

```

octoprint_mrbeam/iobeam/iobeam_handler.py
```diff
@@ -75,7 +75,7 @@ class IoBeamHandler(object):
     # How to get debug info:
     #       echo "info" |  nc -U -w1 /var/run/mrbeam_iobeam.sock
 
-    SOCKET_FILE = "/var/run/mrbeam_iobeam.sock"
+    SOCKET_FILE = "/tmp/mrbeam_iobeam.sock"
```
andywerner, line 78: make this backwards compatible!!!

andywerner (changes requested):

amrsoll:
Closed because of security considerations. It might not be in our best interest to downgrade the privileges of iobeam.

We could reopen again at a later date.

PR #1224, opened 2021-04-01
Fixes #1223 (context menu on mac)

### LINKED ISSUE #1223: Not able to move designs on WA - Happens sometimes
It happens sometimes that a design on the working area is not moveable. It usually does not happen right when you place a new design on the WA, it happens after some action like a (cancelled?) laser job or something.

When in this state the mouse curser does not show a hand, if remains the arrow. (see video)
The issue resolves by reloading browser window or sometimes when the user does some action outside our browser tab.

Here is a video. The user is clicks several times on the selected design to move it.
Video-_7_

andywerner:
Jessie just showed me a very similar behaviour which is technically different. Still need to keep the eyes open...

### DIFF
--- a/octoprint_mrbeam/static/js/mother_viewmodel.js
+++ b/octoprint_mrbeam/static/js/mother_viewmodel.js
@@ -195,6 +195,9 @@ $(function () {
             $("body").on("keyup", function (event) {
                 self.removeClassesForModifierKeys(event);
             });
+            $("body").on("contextmenu", function (event) {
+                self.removeClassesForModifierKeys(event);
+            });
 
             self.addClassesForModifierKeys = function (event) {
                 if (event.which === 16) document.body.classList.add("shiftKey");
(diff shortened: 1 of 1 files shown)


octoprint_mrbeam/static/js/mother_viewmodel.js
```diff
@@ -205,7 +208,16 @@ $(function () {
             self.removeClassesForModifierKeys = function (event) {
                 if (event.which === 16)
                     document.body.classList.remove("shiftKey");
-                if (self.KEY_CODES_CTRL_COMMAND.includes(event.which))
+                if (
+                    self.KEY_CODES_CTRL_COMMAND.includes(event.which) ||
+                    event.type == "contextmenu"
+                )
```
amrsoll, line 214: I think it is best if the "contextmenu" events have the same effect as "keyup" for all the keys -> 
```js
$("body").on("contextmenu", function (event) {
    ["shiftKey", "altKey", "ctrlKey"].foreach(document.body.classList.remove);
});
```

amrsoll (changes requested):
It would be best to make the contextmenu event remove all the key-related identifiers for the body

octoprint_mrbeam/static/js/mother_viewmodel.js
andywerner: as far as we know, this issue is very specific to mac and how the ctrl key opens a context menu.

PR #1228, opened 2021-04-08
Object height pane

andywerner:
Screenshot 2021-04-08 at 22 55 56


khaledsherkawi (approved):

hungerpirat:
Technical feedback: 
- mrbeam.css should be only used for customizing bootstrap classes according to our mr beam style. Styles for these elements should be in tab_workingarea.css. Though mrbeam.css is a mess, it does not get better, if we still add more stuff to it.

Conceptional feedback:
- I miss the clear, visual link between preview and the object height. 
- "surface above the bottom" is imho not clear enough. 
- Jessies suggestion to have a checkbox "Cutting mat in use" did not make it? 
Just my humble two cents. @Gallore has to decide.

andywerner:
@khaledsherkawi looks like commit 989ba42 has an incorrect commit message (I probably reused the one from the commit before by accident...)

PR #1235, opened 2021-04-14
Prevents ready_to_laser if cpu is high (longer waiting time)

andywerner:


octoprint_mrbeam/iobeam/onebutton_handler.py
```diff
@@ -401,36 +401,36 @@ def set_ready_to_laser(self, gcode_file=None):
...
                 cpu_load = sum(map(lambda x: float(x), find_cpu_load.findall(ps_out)))
             except Exception:
                 self._logger.exception("Exception while reading cpu load: ")
                 time.sleep(check_interval)
             self._logger.info(
-                "wait_for_cpu_below() cpu_load: %s, code: %s", cpu_load, code
+                "wait_for_cpu_below() cpu_load: %s, cpu_max: %s, cpu_load > cpu_max: %s",
+                cpu_load,
+                cpu_max,
+                cpu_load > cpu_max,
```
amrsoll, line 433: Please make it a debug log and without the function name

amrsoll (changes requested):
So what happened to `psutils`? It would have been easy to install it through the usb keys as well.

octoprint_mrbeam/iobeam/onebutton_handler.py
andywerner: ok

andywerner:
PSutil:
https://psutil.readthedocs.io/en/latest/
https://pypi.org/project/psutil/

And this:
https://stackoverflow.com/questions/48761714/how-to-get-the-particular-process-cpu-usage-from-task-manager-in-windows-using-p

amrsoll (approved):

amrsoll:
But why not use psutil then?

PR #1241, opened 2021-04-21
Fix condition to log FrontEnd event [1]

amrsoll:
The only reason why this wasn't a problem was because `analyticsEnabled` does not change before startup completes anyway.


andywerner (changes requested):
As kust discussed, this is PR does NOT improve the system.

PR #1258, opened 2021-05-12
fixes logging issue during conversion error handling.

andywerner:


andywerner:
`svgStr` was removed so this log line resulted in an error and prevented the next one from being logger. The next one would actually be useful in this error case

andywerner:
This could potentially be the/one reason for reports from support lately that conversion fails for users without error message. But this must be in stable since 0.8.0 (release end of Dez 2020). So this error has be out there in stable for a few month - can not explain why it would come up only now...

https://support.mr-beam.org/a/tickets/20726

amrsoll (approved):
Alrighty!

PR #1268, opened 2021-06-04
Fixes #1267: Vectornator (generator info and bugs in handling SVGs)

### LINKED ISSUE #1267: SVGs from Vectornator are not handled correctly
We know this issue with SVGs from Vectornator (https://www.vectornator.io/)
which creates SVG paths which use a plus sign as separator between parameters. Snap.svg can not handle this.
[Archive 5.zip](https://github.com/mrbeam/MrBeamPlugin/files/6599282/Archive.5.zip)
Therefore if the path does not contain any whitespace nor comma but some plus signs, we replace these by commas.

andywerner:
Fixes #1267

### DIFF
--- a/octoprint_mrbeam/static/js/matrix_oven.js
+++ b/octoprint_mrbeam/static/js/matrix_oven.js
@@ -208,10 +208,31 @@ Snap.plugin(function (Snap, Element, Paper, global) {
         }
 
         var arr;
+        var arr_orig;
         var d = path_elem.attr("d");
         d = (d || "").trim();
-        var arr_orig;
-        arr = Snap.parsePathString(d);
+
+        // trying to catch and handle bug in Snap.svg
+        // We know this issue with SVGs from Vectornator (https://www.vectornator.io/)
+        // which creates SVG paths which use a plus sign as separator between parameters. Snap.svg can not handle this.
+        // Therefore if the path does not contain any whitespace nor comma but some plus signs, we replace these by commas.
+        if (d.indexOf(" ") < 0 && d.indexOf(",") < 0 && d.indexOf("+") > 0) {
+            let dOld = d;
+            d = d.replaceAll("+", ",");
+            arr = Snap.parsePathString(d);
+            console.warn(
+                "matrix_oven: Handled potential Snap.svg bug: Path contains no whitespace and no comma but plus signs. " +
+                    "After replacing all '+' by ',' the new path is " +
+                    arr.length +
+                    "\n\noriginal d:\n" +
+                    dOld +
+                    "\n\nfixed d:\n" +
+                    d
+            );
+        } else {
+            arr = Snap.parsePathString(d);
+        }
+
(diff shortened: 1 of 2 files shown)


Gallore:
We have a fix: https://github.com/mrbeam/MrBeamPlugin/pull/1268
should go out with the next beta release

octoprint_mrbeam/static/js/matrix_oven.js
```diff
@@ -208,10 +208,31 @@ Snap.plugin(function (Snap, Element, Paper, global) {
...
+        var arr_orig;
         var d = path_elem.attr("d");
         d = (d || "").trim();
-        var arr_orig;
-        arr = Snap.parsePathString(d);
+
+        // trying to catch and handle bug in Snap.svg
+        // We know this issue with SVGs from Vectornator (https://www.vectornator.io/)
+        // which creates SVG paths which use a plus sign as separator between parameters. Snap.svg can not handle this.
+        // Therefore if the path does not contain any whitespace nor comma but some plus signs, we replace these by commas.
```
hungerpirat, line 218: What about an example: "M129.122+146.728L491.004+152.024L491.004+179.68..." => "M129.122,146.728L491.004,152.024L491.004,179.68..."

hungerpirat (commented):
example of the svg change could help understand why and what has been done to avoid the bug.

octoprint_mrbeam/static/js/matrix_oven.js
andywerner: done

octoprint_mrbeam/static/js/matrix_oven.js
```diff
@@ -208,10 +208,32 @@ Snap.plugin(function (Snap, Element, Paper, global) {
...
+            arr = Snap.parsePathString(d);
+            console.warn(
+                "matrix_oven: Handled potential Snap.svg bug: Path contains no whitespace and no comma but plus signs. " +
+                    "After replacing all '+' by ',' the new path is " +
+                    arr.length +
+                    "\n\noriginal d:\n" +
+                    dOld +
+                    "\n\nfixed d:\n" +
+                    d
+            );
```
amrsoll, line 232: It is not worth it to put the dOld string in the console. You are potentially making a huge dump and it can affect performance for no reason when not in dev mode.

I suggest to just replace the `+` into `,` in place for `d` and have `arr = Snap.parsePathString(d);` outside the `if` statement and simply have a single line warning when replacing the `+` characters.

amrsoll (changes requested):
See comment

octoprint_mrbeam/static/js/matrix_oven.js
andywerner: In the beginning I want this output to be available from the customers. (Remember, we send the output to the backend.) However, if there's too much load on the system, it'll drop the call to the backend.

PR #1284, opened 2021-06-25
SW-78 added migration for s series bug with mount_manager

### LINKED ISSUE #62: BUG: USB-Sticks do not get unmounted during a hard power-off
We discovered a serious issue an (likely) all S-Series devices:
The device switches to constant blue lights and Mr Beam Plugin is not started. 
The folder ~/usb_mount still exists and contains the trigger file for the camera plugin. Therefore the device starts the camera plugin and does NOT start Mr Beam Plugin.

Immediate actions taken:
We created a dummy-usb stick which does almost nothing, but it replaces the trigger file in the ~/usb_mount folder. 

Needed Steps:
- Image needs to call `mount_manager clear` on during boot. This should happen before any potentially plugged in USB stick gets mounted
- Install this also on all S devices ALREADY SHIPPED. (Probably in a migration step)
- Test that with this change it is not possible that a device goes in calibration mode without a usb stick even if the device got hard powered off with the stick still plugged in.
- Make sure that ~/usb_mount folders never exists if no usb stick is plugged in (either after stick got unplugged or after a reboot.) Especially test with a stick wich reboots the device when unplugged. (Josef found an issue where the rebot happened before the mount_manager was able to remove this folder.)

JFSolution:
fix https://github.com/mrbeam/BeamOS/issues/62


amrsoll (changes requested):
As I explained in https://github.com/mrbeam/BeamOS/issues/62#issuecomment-869418704 , this will Not have the intended outcome because the affected devices will not be able to perform the migration.

octoprint_mrbeam/migrate.py
```diff
@@ -214,6 +215,19 @@ def run(self):
                 ):
                     self.hostname_helper_scripts()
 
+                beamos_tier, beamos_date = self.plugin._device_info.get_beamos_version()
+                if beamos_date <= datetime.strptime(
+                    "2021-06-25", "%Y-%m-%d"
+                ).date() and (
+                    self.version_previous is None
```
amrsoll: Also this might raise a None-comparison error if `beamos_date` is None.

Always assume that some info read from a "config" file can be None

JFSolution: the get_beamos_version returns the legacy date if it can't get the configdata, but i added a condition to handle it

JFSolution:
I spoke with @Gallore and he said, we should make this migration, so all devices have the same base, and we don't have to remember this special case.
The devices will be able to migrate because the users already received a usb-stick that hotfixes this issue

octoprint_mrbeam/migrate.py
```diff
@@ -627,6 +645,21 @@ def disable_gcode_auto_deletion(self):
...
+        fixes a problem with the images before 25.6.2021
+        the rc.local file was missing the clear command for the mount_manager
+        this replaces the rc.local file with the one containing this row
+        """
+        self._logger.info("start fix_s_series_mount_manager")
+        src = os.path.join(
+            __package_path__, self.MIGRATE_FILES_FOLDER, "fix_s_series_mount_manager"
+        )
+        dst = "/etc/rc.local"
+        exec_cmd("sudo cp {src} {dst}".format(src=src, dst=dst))
```
andywerner, line 659: get the return value (Boolean) and add it to the following line (logging), so that the logged information is always correct.

andywerner (approved):
approved with minor change requests.

octoprint_mrbeam/migrate.py
amrsoll: Or have it as an if statement
```python
if(exec_cmd( ... )):
  log("Great")
else:
  log("not so great")

amrsoll: In fact, `exec_cmd()` already logs an error when the exit code is not `0` so I don't think this is necessary

JFSolution: I did it with the if else

JFSolution:
merged changes to fix that the migration won't start for s series images

JFSolution:
Tested it on my device, the migration works now

JFSolution:
fixed the branch with @amrsoll new PR https://github.com/mrbeam/MrBeamPlugin/pull/1307

PR #1287, opened 2021-06-30
SW-80 cloud config for update

JFSolution:
Changing the update mechanism to look in a json file for infos like the repository, branch, ....
This file will be hostet in the internet and sync at every start of the device or at a tier change


octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
+
+    def look_for_new_config_file(self):
+        """
+        starts a thread to look online for a new config file
+        """
+        th = threading.Thread(target=self._load_update_file_from_cloud)
+        th.setName("MrBeamSoftwareupdateHandler:_load_update_file_from_cloud")
+        th.daemon = True
+        th.start()
+        th.join()
```
andywerner: please remove the join

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
+
+    def _load_update_file_from_cloud(self):
+        """
+        overrides the local update config file if there is a newer one on the server
+        """
+        _logger.debug("load update file")
+        newfile = False
+        try:
+            r = requests.get(
```
andywerner: what happens if server returns an error (400, 500, ...?)

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
+                                update_info["version"],
+                            )
+
+                    except ValueError:
+                        newfile = True
+                        _logger.error(
+                            "there is a wrong configured config local file - override local file with server file"
+                        )
+                    finally:
+                        # override local file
```
andywerner: Please do now re-write the file if it is the same version

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
+        starts a thread to look online for a new config file
+        """
+        th = threading.Thread(target=self._load_update_file_from_cloud)
+        th.setName("MrBeamSoftwareupdateHandler:_load_update_file_from_cloud")
+        th.daemon = True
+        th.start()
+        th.join()
+
+    def _load_update_file_from_cloud(self):
+        """
```
andywerner, line 71: Because this runs in a separate thread, please add a try/catch all and log the exception at the top most level of the thread.

octoprint_mrbeam/static/updates/update_info.json
```diff
@@ -0,0 +1,123 @@
+{
```
andywerner: this should be removed

andywerner: We need to make sure that the SW always shows the currently installed versions of the SW components - even if we do not have a locally downloaded version and no internet connection.

andywerner (changes requested):
We need to make sure that the SW always shows the currently installed versions of the SW components - even if we do not have a locally downloaded version and no internet connection.

octoprint_mrbeam/software_update_information.py
JFSolution: added exception handling for this case

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
+        self._beamos_date = beamos_date
+
+    def look_for_new_config_file(self):
+        """
+        starts a thread to look online for a new config file
+        """
+        th = threading.Thread(target=self._load_update_file_from_cloud)
+        th.setName("MrBeamSoftwareupdateHandler:_load_update_file_from_cloud")
+        th.daemon = True
+        th.start()
```
amrsoll: 1. If you want to use a separate thread, please use the `get_thread` decorator: it automatically sets the name and logs any uncaught exception. If the Only purpose of this function is to create a Thread, then you can just decorate `_load_update_file_from_cloud` with the `@get_thread(daemon=True)` decorator, and you won't need to keep `look_for_new_config_file()`
2. What is the advantage of creating a separate thread here if you are going to join it directly? It looks like this can be done synchronously here.

octoprint_mrbeam/software_update_information.py
amrsoll: See my previous comment

JFSolution: ok

octoprint_mrbeam/static/updates/update_info.json
JFSolution: ok, it will fallback to mrbeamplugin file if there is no from server or localy

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
+                                "update local file from server - %s -> %s",
+                                serverfile_info["version"],
+                                update_info["version"],
+                            )
+
+                    except ValueError:
+                        newfile = True
+                        _logger.error(
+                            "there is a wrong configured config local file - override local file with server file"
+                        )
```
amrsoll: When it comes to un-sanitised content like this JSON info, I would suggest to use `dict.get()`
```python
server_version = serverfile_info.get("version", None)
update_version = update_info.get("version", None)
if server_version is not None:
  ...
```
This way you 
1. don't have to try/except the `valueError` if the version is not available 
2. Can decide what to use when `version` is not available, either from the server or from the update_info.

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+    @param _softwareupdate_handler: the handler class to look for a new config file online
+    """
+    try:
+        with open(join(plugin._settings.getBaseFolder("base"), SW_UPDATE_FILE)) as f:
+            update_info = json.load(f)
+    except IOError as e:
+        _softwareupdate_handler.look_for_new_config_file()
+        _logger.error(
+            "could not find file %s - error: %s",
+            join(plugin._settings.getBaseFolder("base"), SW_UPDATE_FILE),
```
amrsoll, line 324: Please put this path into a function if you need to create it dynamically. You are using the same line too many times. (Rule of thumb: do not repeat yourself more than 3 times)

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+            e,
+        )
+    except ValueError as e:
+        _softwareupdate_handler.look_for_new_config_file()
+        _logger.error(
+            "there is a wrong configured config file local - try to load from server"
+        )
+        _logger.debug("error %s - %s", e, f)
```
amrsoll, line 332: `f` might not be available here.
Better to log the path instead of the file object as well.

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
     else:
```
amrsoll, line 333: This `else` statement goes aaaall the way to the end of the function. Please return when the exceptions happen instead of indenting 130 lines.

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
     )
 
-
-def _set_info_mrb_hw_info(plugin, tier, beamos_date):
-    if beamos_date > BEAMOS_LEGACY_DATE:
-        pip_command = "sudo /usr/local/iobeam/venv/bin/pip"
+    # plugin._settings.global_set_boolean(op_swu_keys + ["prerelease"], False)
+
+
+def _set_info_from_file(plugin, tier, beamos_date, _softwareupdate_handler):
```
amrsoll, line 294: This function is too big. I would recommend to split it up into multiple functions with good descriptive names of what they are doing.

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+                            "displayVersion": current_version,
+                        }
+                    )
+            except:
+                _logger.error("Error in module %s %s", module_id, sys.exc_info()[0])
+        _logger.debug(sw_update_config)
+
+
+def _set_update_config_from_dict(update_config, dict):
+    update_config.update(dict)
```
amrsoll, line 465: 1. This function on the other hand is too small to make any sense.
2. Did you mean to make a shallow merge? Otherwise, please use the `octoprint.util.dict_merge()` function to make a deep merge of the tree structure

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+
+
+def _get_display_name(plugin, name):
+    return name
```
amrsoll, line 469: Same here, Did you mean to add more content to this function for this PR?
If yes, then please set the PR as draft until the changes are reflected
If not, then please do not add extra functions for potential future code, because then it gets lost.

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+    tierversion = get_tier_by_id(tier)
+    if tierversion in config:
+        module_tier = config[tierversion]
+        _set_update_config_from_dict(
+            config, module_tier
+        )  # set tier config from default settings
+
+    if "beamos_date" in config:
+        beamos_date_config = config["beamos_date"]
+        for date, beamos_config in beamos_date_config.items():
```
amrsoll, line 267: Tip: you can shorten this to
```python
for date, beamos_config in config.get("beamos_date" ,{}).items():
    ...
```

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+                    if tierversion in moduleconfig:
+                        module_tier = moduleconfig[tierversion]
+                        _set_update_config_from_dict(
+                            moduleconfig, module_tier
+                        )  # set tier config from default settings
+                    if tierversion in module:
+                        module_tier = module[tierversion]
+                        _set_update_config_from_dict(
+                            moduleconfig, module_tier
+                        )  # override tier config from tiers set in config_file
```
amrsoll, line 373: the Black formatter does not like comments at the end of lines

Better to have them in a separate line (the previous line) so they don't create these ugly parenthesis-newline patterns

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+
+                    _set_update_config_from_dict(
+                        moduleconfig, module
+                    )  # set default config from file for module
+
+                    # have to be after the default config from file
+                    if "beamos_date" in module:
+                        beamos_date_config = module["beamos_date"]
+                        for date, beamos_config in beamos_date_config.items():
```
amrsoll, line 382: My previous tip also works here :)

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+                            _logger.debug(
+                                "date compare %s >= %s -> %s",
+                                beamos_date,
+                                datetime.strptime(date, "%Y-%m-%d").date(),
+                                beamos_config,
+                            )
```
amrsoll, line 388: Please do not log if conditions, that's too much...

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+                                if tierversion in beamos_config:
+                                    beamos_config_module_tier = beamos_config[
+                                        tierversion
+                                    ]
+                                    _set_update_config_from_dict(
+                                        beamos_config, beamos_config_module_tier
+                                    )  # override tier config from tiers set in config_file
+                                _set_update_config_from_dict(
+                                    moduleconfig, beamos_config
+                                )
```
amrsoll, line 402: This is where a deep merge using `dict_merge` is nice; I think it can replace this extra `if` statement

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+                        moduleconfig["pip_command"] = GLOBAL_PIP_COMMAND
+                    if "pip_command" in moduleconfig:
+                        pip_command = moduleconfig["pip_command"]
+                        # if global_pip_command is set module is installed outside of our virtualenv therefor we can't use default pip command.
+                        # /usr/local/lib/python2.7/dist-packages must be writable for pi user otherwise OctoPrint won't accept this as a valid pip command
+                        # pip_command = GLOBAL_PIP_COMMAND
+                        package_name = (
+                            module["package_name"]
+                            if "package_name" in module
+                            else module_id
```
amrsoll, line 430:
```python
package_name = module.get("package_name", module_id)
```

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
+                        module,
+                        moduleconfig,
+                        "pip_command" not in moduleconfig,
+                    )
+                    if (
+                        "global_pip_command" in module
+                        and "pip_command" not in moduleconfig
+                    ):
+                        moduleconfig["pip_command"] = GLOBAL_PIP_COMMAND
+                    if "pip_command" in moduleconfig:
```
amrsoll, line 422: This is more for readability, but I guess this `if` could be an `elif` ? It helps with understanding what we are doing.

octoprint_mrbeam/software_update_information.py
```diff
@@ -100,231 +215,244 @@ def switch_software_channel(plugin, channel):
...
-        branch=branch,
-        branch_default=branch_default,
-        restart=restart,
-        **kwargs
-    )
-    return {module_id: update_info}
+        try:
+            _set_octoprint_config(
+                plugin, tier, update_info.pop("octoprint"), beamos_date
+            )
```
amrsoll: Is there a reason why you want to `.pop()` those values instead of `.get()` ? Couldn't it lead to unpredicted missing info later on?

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
 GLOBAL_PIP_BIN = "/usr/local/bin/pip"
 GLOBAL_PIP_COMMAND = (
     "sudo {}".format(GLOBAL_PIP_BIN) if os.path.isfile(GLOBAL_PIP_BIN) else None
 )
-# GLOBAL_PIP_COMMAND = "sudo {} -m pip".format(GLOBAL_PY_BIN) if os.path.isfile(GLOBAL_PY_BIN) else None #  --disable-pip-version-check
-# VENV_PIP_COMMAND = ("%s -m pip --disable-pip-version-check" % VENV_PY_BIN).split(' ') if os.path.isfile(VENV_PY_BIN) else None
+
 BEAMOS_LEGACY_DATE = date(2018, 1, 12)
 
+_softwareupdate_handler = None
```
amrsoll, line 45: Why are we using this as a global?

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
 
+def get_modules():
+    return sw_update_config
+
+
+class MrBeamSoftwareupdateHandler:
+    def __init__(self, plugin, tier, beamos_date):
+        self._plugin = plugin
+        self._tier = tier
+        self._beamos_date = beamos_date
```
amrsoll, line 56: Why is this class necessary?
It does not seem like you are using it's attributes, and `self` is only used for referencing `self._plugin`, so to me it looks redundant. I would suggest to keep this on a functional approach instead of being Object Oriented.

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
+SW_UPDATE_CLOUD_PATH = "https://mr-beam.org/beamos/config/sw-update-conf"
+
 # add to the display name to modules that should be shown at the top of the list
 SORT_UP_PREFIX = " "
 
-
 _logger = mrb_logger("octoprint.plugins.mrbeam.software_update_information")
 
+sw_update_config = dict()
+
```
amrsoll, line 34: Please do NOT use mutable globals as much as possible.

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,69 +1,178 @@
...
-                _set_info_mrb_hw_info(plugin, tier, beamos_date),
-                # _config_octoprint(plugin, tier),
-            ],
-        )
-    )
+    _logger.info("SoftwareUpdate using tier: %s %s", tier, beamos_date)
+    global _softwareupdate_handler
+    if _softwareupdate_handler is None:
+        _softwareupdate_handler = MrBeamSoftwareupdateHandler(plugin, tier, beamos_date)
+        _softwareupdate_handler.look_for_new_config_file()
```
amrsoll, line 191: If you decide to remove the `MrBeamSoftwareupdateHandler` class, then you can just pass the plugin directly to `look_for_new_config_file(plugin)`.

amrsoll (changes requested):
Hey, I would like to have some of my concerns addressed if possible.

I had refactored this part to be more function-based with immutable variables. I do not think this is the right approach to solve the problem.

octoprint_mrbeam/software_update_information.py
amrsoll: N.B. when a function is decorated with `get_thread`, it will return a `Thread` object which you can decide to `.join()` later on

JFSolution: the join was not intended, the thread was requested from andy so the eventhandler will not be blocked by this call

octoprint_mrbeam/software_update_information.py
JFSolution: ok this looks neat, I will try this

octoprint_mrbeam/software_update_information.py
JFSolution: ok will refactor this

octoprint_mrbeam/software_update_information.py
Josef-MrBeam: ok will change

octoprint_mrbeam/software_update_information.py
JFSolution: I pop it so it wont be handeld by the for loop that processes all the module infos

octoprint_mrbeam/software_update_information.py
JFSolution: To make sure that there are not multiple calls that run at the same time

octoprint_mrbeam/software_update_information.py
amrsoll: In that case I would say that they shouldn't be linked to the same node if they are not going to be treated the same; so we don't have to do risky things like this.

octoprint_mrbeam/software_update_information.py
amrsoll: Then I would suggest to use a `Lock` / `Rlock` instead :)

octoprint_mrbeam/software_update_information.py
JFSolution: hmm, I can't remember why I choose to do it Object oriented, you are right I only use the self._plugin
What is the downside to do it this way?

octoprint_mrbeam/software_update_information.py
JFSolution: I looked it up I used it as it was before your buster changes, I can change this like you did for the buster image

octoprint_mrbeam/software_update_information.py
JFSolution: ok, changed

octoprint_mrbeam/software_update_information.py
JFSolution: shouldn't it be without the daemon, this could cause file corruption when it gets killed while writing

octoprint_mrbeam/software_update_information.py
JFSolution: the idea was to log what is the problem in the file and it should be available, because its just a valueerror of the json.load

octoprint_mrbeam/software_update_information.py
JFSolution: Ok I can put it in a separate node, but why it is risky to pop it? I know that I don't need this data later

JFSolution:
new PR https://github.com/mrbeam/MrBeamPlugin/pull/1389

PR #1288, opened 2021-07-01
Environment selection added

khaledsherkawi:


octoprint_mrbeam/static/js/settings/dev_design_store.js
```diff
@@ -57,6 +59,17 @@ $(function () {
             self.showAuthToken();
         };
 
+        self.changeEnv = function () {
+            if (self.selectedEnv() === 'prod') {
+                self.designStore.DESIGN_STORE_IFRAME_SRC = 'https://designs.cloud.mr-beam.org';
```
andywerner: it's a good practice to define data like URLs as constants at the beginning of the file.

andywerner (approved):

PR #1289, opened 2021-07-05
SW-113 Design Store necessary modifications 

khaledsherkawi:


octoprint_mrbeam/static/js/design_store.js
```diff
@@ -114,6 +132,15 @@ $(function () {
             self.saveTokenInUserSettings(payload.token);
         };
 
+        self.onLastUploadedDateReceived = function (payload) {
+            let oldLastUploaded = self.getLastUploadedDate();
+            if (payload.last_uploaded !== "" && oldLastUploaded !== payload.last_uploaded) {
```
andywerner: Question: Why do you to a type-safe check for not empty sting? `self.getLastUploadedDate()` could also return `undefined` which would be uneuqal to `payload.last_uploaded` and could therefor trigger to red-dot to appear. (That's how I read this.) 
I suggest to just check `payload.last_uploaded` for a true-ish value

```suggestion
            if (payload.last_uploaded && oldLastUploaded !== payload.last_uploaded) {
```

andywerner (changes requested):
See line comment.
Also: please append a screenshot to the PR.

khaledsherkawi:
Screenshot from 2021-07-05 14-54-59

PR #1290, opened 2021-07-06
SW-136 Vector engraving getting skipped when min_intensity greater than 0

amrsoll:
* Introduce ``observableInt()`` which forces a shadow `ko.observable` to be an integer.


ahmed-mrbeam:
@amrsoll  would you please add the issue ID to the PR name ?

amrsoll:
It is all in Jira, I'll add the ticket ID to the title too

JFSolution (approved):

PR #1299, opened 2021-07-12
Using redirected URL for laser safety confirmation upload

andywerner:
Still requires an actual test!


amrsoll:
Converted to draft as long as the `.htaccess` file was not updated

PR #1300, opened 2021-07-14
SW-171: OK, next try: Don't do any passive logins during boot phase (and 10 secs after), also wait 3 secs before actually performing the passive login.

andywerner:


octoprint_mrbeam/static/js/mrbeam.js
```diff
@@ -440,10 +443,17 @@ $(function () {
         };
 
         self._handle_session_expired = function (triggerUrl) {
+            // don't do this during boot time.
+            if (
+                self.isCurtainOpened <= 0 &&
+                self.isCurtainOpened + 10 * 1000 > Date.now()
```
khaledsherkawi, line 449: is 10 seconds here an assumption for how long the booting takes??

khaledsherkawi (dismissed):

octoprint_mrbeam/static/js/mrbeam.js
khaledsherkawi: I'm not sure the "less than" services any purpose here. please clarify.

andywerner: 1. It is the moment the "onCurtainOpened" event fired + 10 seconds. 
2. self.isCurtainOpened is initialised to 0 (I see it as a timestamp). It remains zero until the "onCurtainOpened" event fires, then it is set ti the current timestamp. The "less than" is "just to be on the save side" incase someone initialises or sets it's value to -1.

khaledsherkawi (approved):

octoprint_mrbeam/static/js/mrbeam.js
amrsoll: wouldn't it be easier to check if the curtain is still active or displayed?

PR #1301, opened 2021-07-14
Feature/sw 155 test add user

sherifhussien:


amrsoll (approved):
This is looking good :) I will ask Josef to review it as well on his mac

octoprint_mrbeam/util/uptime.py
```diff
@@ -1,14 +1,19 @@
...
 from octoprint_mrbeam.mrb_logger import mrb_logger
 
 # http://planzero.org/blog/2012/01/26/system_uptime_in_python,_a_better_way
 def get_uptime():
     try:
-        with open("/proc/uptime", "r") as f:
-            uptime = float(f.readline().split()[0])
-        return uptime
+        if platform == 'darwin':
+            raise NotImplementedError('uptime not implemented for darwin')
```
andywerner: on Mac, you can just call `uptime`. 
The resulting output has this format: 
> 10:33 up 21:40, 4 users, load averages: 0.09 0.19 0.21

( https://osxdaily.com/2009/09/22/check-your-macs-uptime-and-reboot-history/ ) 

or use `sysctl -n kern.boottime | cut -c14-18`
which return just an int for the seconds:
> 29101

sherifhussien: can we use this instead of handling each manually? https://pythonhosted.org/uptime/ @Gallore

octoprint_mrbeam/util/uptime.py
```diff
@@ -1,14 +1,19 @@
...
 
 # http://planzero.org/blog/2012/01/26/system_uptime_in_python,_a_better_way
 def get_uptime():
     try:
-        with open("/proc/uptime", "r") as f:
-            uptime = float(f.readline().split()[0])
-        return uptime
+        if platform == 'darwin':
+            p = os.popen("uptime")
+            return p.read()
```
amrsoll: Where does this come from? Is it a file??

octoprint_mrbeam/util/uptime.py
amrsoll: For technical reasons, the uptime package is not installed on our devices and will not be installed yet. You can add the `uptime` package in `setup.py` as an optional requirements for `[develop]`

octoprint_mrbeam/util/uptime.py
sherifhussien: No, it's just a command on Mac

https://github.com/mrbeam/MrBeamPlugin/pull/1301#discussion_r681002957

amrsoll: Ah ok, then `os.system()` is a more common way of handling a shell command

amrsoll: I would be careful with `os.popen` because I think you have to handle a file descriptor (in this case it's `p`)

sherifhussien: changed

amrsoll (approved):

octoprint_mrbeam/util/uptime.py
andywerner: I'm fine with either way.

PR #1303, opened 2021-07-16
SW-207 fix migration for s series

JFSolution:
fixes the migration so it will run for the s-series


amrsoll (approved):
Alright :) I see the PR is towards your other branch for the S series migration :+1:

PR #1307, opened 2021-07-19
SW-78 added migration for s series bug with mount manager

amrsoll:


JFSolution:
have to test it again, after commit cleanup

amrsoll (commented):
@Josef-MrBeam 

It would seem that `octoprint_mrbeam/files/migrate/fix_s_series_mount_manager` is no longer needed.

Also the name for the updated mount_manager is misleading. It would be best to overwrite the old mount_manager in the migration files instead of adding it as a new file. It should not affect the migration if you use the old file path.

Please also rebase the branche(s) on the latest develop - there seems to be extra diffs in this PR that are not expected.

In order to fix the root cause, I would also like to see the version being written into `config.yaml` regardless of whether the migration is performed or not.

Given that I am the author of the PR, GitHub doesn't allow me to mark "Request Changes", but that is my feedback.

JFSolution:
> It would seem that `octoprint_mrbeam/files/migrate/fix_s_series_mount_manager` is no longer needed.

it is needed because the image of the 28.6. already contains the change to the rc.local file so this will make sure that all are on the same stat of this file 

> Also the name for the updated mount_manager is misleading. It would be best to overwrite the old mount_manager in the migration files instead of adding it as a new file. It should not affect the migration if you use the old file path.

I thought about that and there were so many changes done to this file, that I thought this is needed for some of the devices, and it could be that some of theme still needs this change, so I thought it would be best to keep this file

> Please also rebase the branche(s) on the latest develop - there seems to be extra diffs in this PR that are not expected.

I did the rebase, the changed files are just from this pullrequest

> In order to fix the root cause, I would also like to see the version being written into `config.yaml` regardless of whether the migration is performed or not.

okay I can add this

amrsoll:
> I did the rebase, the changed files are just from this pullrequest

It is not normal that you have extra reformating


For the new `mount_manager` file; if it is not targeted for the legacy images, then it is best to keep them separated so we do not update the file for the legacy images :+1:

amrsoll:
And please rename `fix_s_series_mount_manager` to `rc.local` - we will probably be using it for the legacy images as well later.

amrsoll:
Please check your `black` version in the virtualenv
```sh
$ black --version
black, version 20.8b1
```
If there is a mismatch:
```sh
pip install black==20.8b1
```
and redo the pre-commit on those files

octoprint_mrbeam/migrate.py
```diff
@@ -627,6 +652,56 @@ def disable_gcode_auto_deletion(self):
...
+        src_rc_local = os.path.join(
+            __package_path__, self.MIGRATE_FILES_FOLDER, "mount_manager.rules"
+        )
+        dst_rc_local = "/lib/udev/rules.d/00-mount_manager.rules"
+        if exec_cmd("sudo cp {src} {dst}".format(src=src_rc_local, dst=dst_rc_local)):
+            self._logger.info("updated mountmanager udev rules", dst_rc_local)
+            exec_cmd("sudo rm /etc/systemd/system/usb_mount_manager_add.service")
+            exec_cmd("sudo rm /etc/systemd/system/usb_mount_manager_remove.service")
+        self._logger.info("end fix_s_series_mount_manager")
+        self.reboot_needed = True
```
amrsoll: Why do we need a reboot though?

JFSolution: Yes we need this, it only work for me after the reboot

octoprint_mrbeam/migrate.py
```diff
@@ -214,24 +217,48 @@ def run(self):
...
 
-                self.save_current_version()
                 self._logger.info(
                     "Finished migration from v{} to v{}.".format(
                         self.version_previous, self.version_current
                     )
                 )
+                if self.reboot_needed:
+                    self._logger.info("reboot needed, will reboot now")
+                    exec_cmd("sudo reboot now")
```
amrsoll: We need to write the version **before** the reboot

JFSolution:
> Please check your `black` version in the virtualenv
> 
> ```shell
> $ black --version
> black, version 20.8b1
> ```
> 
> If there is a mismatch:
> 
> ```shell
> pip install black==20.8b1
> ```
> 
> and redo the pre-commit on those files

I had the version 20.8b1 installed

octoprint_mrbeam/migrate.py
JFSolution: ohh yes, fixed that

octoprint_mrbeam/migrate.py
```diff
@@ -627,6 +652,56 @@ def disable_gcode_auto_deletion(self):
...
+                success = False
+            self._logger.info("enable %s", systemdfile)
+            if not exec_cmd("sudo systemctl daemon-reload"):
+                success = False
+            if not exec_cmd(
+                "sudo systemctl enable {systemdfile}".format(systemdfile=systemdfile)
+            ):
+                success = False
+            if success:
+                self._logger.info("successfully created ", systemdfile)
```
amrsoll: This is a little python hack:
```python
success &= exec_cmd(...)
# Same as
success = success and exec_cmd(...)
```
:)
But you do not seem to do anything if it is not successful?

JFSolution: Ahh I was looking for something like that :)

octoprint_mrbeam/migrate.py
amrsoll: What works? Perhaps you can also disable those systemd services before removing?

octoprint_mrbeam/migrate.py
amrsoll:
```python
if exec_cmd("sudo cp {src} {dst}".format(src=src, dst=dst))
   and exec_cmd("sudo systemctl daemon-reload")
   and exec_cmd("sudo systemctl enable {systemdfile}".format(systemdfile=systemdfile)
):
    self._logger.info("successfully created ", systemdfile)
```

amrsoll:
```python
systemdfiles = (
    (True, "mount_manager_remove.service"),
    (True, "mount_manager_remove_before_octo.service"),
    (False, "mount_manager_add.service"),
)
for enable, systemdfile in systemdfiles:
  ...

Comment (amrsoll) [reply octoprint_mrbeam/migrate.py]:
```python
if exec_cmd("sudo cp {src} {dst}".format(src=src, dst=dst))
   and exec_cmd("sudo systemctl daemon-reload")
   and (not enable or exec_cmd("sudo systemctl enable {}".format(systemdfile))
):
    self._logger.info("successfully created ", systemdfile)
```

amrsoll:
All issues resolved :)

JFSolution (approved):
approved by Axel and me

PR #1310, opened 2021-07-23
SW-172 Fix settings selection

khaledsherkawi:
- Add 'active' class to first element only


JFSolution (approved):
approved

amrsoll:
Thank you, I would have kept this PR open for now - I would have merged it onto the new branch...
But ok this way I see the commit in the reference branch

PR #1316, opened 2021-07-31
SW-180 Design Store: guest mode is shown in Mr Beam UI

khaledsherkawi:
Conditions added to avoid console errors upon logging in with newly created users

### DIFF
--- a/octoprint_mrbeam/static/js/design_store.js
+++ b/octoprint_mrbeam/static/js/design_store.js
@@ -23,8 +23,8 @@ $(function () {
 
         self.getEmail = function () {
             if (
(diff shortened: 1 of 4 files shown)


andywerner (changes requested):
This OptionalChain operator is only supported since Chrome 80. Please also change the min version of Chrome from 60 to 80 here: https://github.com/mrbeam/MrBeamPlugin/blob/2fd4065bffeb430b6eddb09ee7228c2567237649/octoprint_mrbeam/static/js/mrbeam.js#L42

andywerner (approved):

PR #1317, opened 2021-08-05
Bugfix/sw 267 logged out after login closed loop

khaledsherkawi:


octoprint_mrbeam/static/js/material_settings.js
```diff
@@ -43,16 +46,18 @@ $(function () {
...
                 })
                 .fail(function (response) {
-                    setTimeout(
-                        self.loadMaterialSettings,
-                        self.MATERIAL_SETTINGS_RETRY_TIME,
-                        callback
-                    );
-                    console.error(
-                        "Unable to load material settings. Retrying in " +
+                    if (self.loginState.loggedIn()) {
```
andywerner, line 49: Please add an else statement and log a message like "Unable to load material settings. Not retrying because user is not logged in."

octoprint_mrbeam/static/js/messages.js
```diff
@@ -19,6 +19,11 @@ $(function () {
         self.notificationsHandled = false;
         self.messagesLoaded = false;
 
+        self.onStartupComplete = function () {
+            // Hide Messaging icon
+            $("li a#messages_nav_tab").hide();
```
andywerner, line 24: I assume that `onUserLoggedIn` is called before `onStartupComplete` in the boot order. I assume you tested it and it worked, so maybe practice beats theory. Still, I suggest to put the hiding in `onStartup`. For this command you do not the the viewmodel to be bound therefore onStartup should be fine.

octoprint_mrbeam/static/js/mrbeam.js
```diff
@@ -443,6 +445,24 @@ $(function () {
         };
 
         self._handle_session_expired = function (triggerUrl) {
+            if (self.isCurtainOpened > 0 && self.triggerAnalytics) {
```
andywerner: Please change as we just discussed: Count 401 errors for 10 sec and then submit the result in 1 request to the backend.

octoprint_mrbeam/static/js/settings/camera_settings.js
```diff
@@ -235,7 +235,7 @@ $(function () {
                                         response.responseText
                                 ),
                                 type: "warning",
-                                hide: false,
+                                hide: true,
```
andywerner, line 238: :-)

octoprint_mrbeam/static/js/mrbeam.js
```diff
@@ -499,9 +519,14 @@ $(function () {
                             // give it some time to settle in an endless loop
                             setTimeout(function () {
                                 OctoPrint.socket.reconnect();
+                                // Add to analytics to check how often passive logins are used after 401 errors
+                                self.analytics.send_fontend_event(
+                                    "passive_login",
```
andywerner, line 530: Since you add new events, please increase the version number of analytics in analytics handler and also add these to this documentation: https://docs.google.com/document/d/1G2UETo65dsFy_X6OhoSsyo0XIFSgjfoFy-n-5RSYJPg/edit

andywerner (changes requested):

octoprint_mrbeam/static/js/mrbeam.js
```diff
@@ -443,6 +445,22 @@ $(function () {
...
+                        let error401Count = self.error401Count;
+                        let payload = {
+                            error401Count: error401Count,
+                        };
+                        self.analytics.send_fontend_event(
+                            "expired_session",
+                            payload
+                        );
+                        self.error401Count = 0;
+                    }, 10000);
```
andywerner: Please set the timeout to 2000.
```suggestion
                    }, 2000);
```

octoprint_mrbeam/static/js/mrbeam.js
```diff
@@ -443,6 +445,22 @@ $(function () {
         };
 
         self._handle_session_expired = function (triggerUrl) {
+            if (self.isCurtainOpened > 0) {
+                self.error401Count++;
+                if (self.error401Count === 1) {
+                    setTimeout(() => {
+                        let error401Count = self.error401Count;
```
andywerner, line 457: Minor details: I don't see a reason to create this local variable. Further down you could read directly from self.error401Count

andywerner: it is important

andywerner (changes requested):

andywerner (approved):

PR #1320, opened 2021-08-06
SW 265 font ligatures css set to none

khaledsherkawi:


octoprint_mrbeam/static/js/working_area.js
```diff
@@ -3881,15 +3893,25 @@ $(function () {
...
 
-            var text = uc.text(0, 0, placeholderText);
-            text.attr({
+            const curvedText = uc.text(0, 0, placeholderText);
+            curvedText.attr({
                 style:
                     "white-space: pre; font-size: " +
                     size +
-                    "px; font-family: Ubuntu; text-anchor: middle",
+                    "px; font-family: Ubuntu; text-anchor: middle; font-variant-ligatures: none;",
```
khaledsherkawi, line 3901: For Ligatures to be shown on curved texts this extra CSS should be removed. Is there any reason to leave it?

hungerpirat: Ligatures can't be shown on curved text. 
Therefore they're always disabled here. 
This solution has two nodes (text.straightText, text.curvedText), the latter one has ligatures permanently disabled.

hungerpirat (approved):

PR #1322, opened 2021-08-10
SW-192 Only enable terminal for DEV environement

sherifhussien:


amrsoll (approved):

amrsoll (changes requested):
Actually it doesn't work, you should only see the terminal tab when in dev mode.

So you have to make sure that the setting reverts to False if you are not in a dev environment.

See https://github.com/mrbeam/MrBeamPlugin/pull/1322/commits/7f7fedf7bbdb3a0f3da1fcfab7bc57bdb5956da5

`terminalEnabled=self._settings.get(["terminal"]) and self.is_dev_env() or self.support_mode,`

The terminal should automatically be enabled in support mode (I can show you) but not in dev mode

sherifhussien:
> Actually it doesn't work, you should only see the terminal tab when in dev mode.

@amrsoll 
Based on  Andy's comment on the story, that shouldn't be the case

https://mr-beam.atlassian.net/browse/SW-192?focusedCommentId=10768

amrsoll (approved):

PR #1323, opened 2021-08-13
SW-376: Defective serials in list of sold devices (managed by Lena)

andywerner:
…actual serial number.


ahmed-mrbeam:
tested using the online tool:
http://labelary.com/viewer.html and now we don't have the missing leading zero after scanning the QR code

ahmed-mrbeam (approved):

PR #1325, opened 2021-08-16
Feature/SW-340 Add EAN numbers for DCS single + bundle.

amrsoll:
To keep the changes minimal, I had decided to hard code them despite the ticket requirements. 
1. It wouldn't make sense to add new logic to load from a config file unless we are completely refactoring/migrating into a new project.
2. We don't know which config file to read from / create config file where?

### DIFF
--- a/octoprint_mrbeam/camera/label_printer.py
+++ b/octoprint_mrbeam/camera/label_printer.py
@@ -32,12 +32,14 @@ class LabelPrinter(object):
         MRBEAM2_DC_R1=None,
         MRBEAM2_DC_R2=None,
         MRBEAM2_DC=dict(single="4260625360156", bundle="4260625360163"),
+        MRBEAM2_DC_S=dict(single="4260625361023", bundle="4260625361030"),
     )
 
     MRBEAM_2 = "MRBEAM2"
     MRBEAM_2_DC_R1 = "MRBEAM2_DC_R1"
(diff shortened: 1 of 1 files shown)


andywerner (approved):

octoprint_mrbeam/camera/label_printer.py
```diff
@@ -221,6 +214,8 @@ def _get_model_abbrev(self):
             model = "DCR"
         elif self._device_info.get_model() in (self.MRBEAM_2_DC):
             model = "DC"
+        elif self._device_info.get_model() == self.MRBEAM_2_DC_S:
```
khaledsherkawi, line 217:
```suggestion
        elif self._device_info.get_model() in self.MRBEAM_2_DC_S:
```
just for consistency

khaledsherkawi (approved):

octoprint_mrbeam/camera/label_printer.py
amrsoll: so you mean `X in (self.MRBEAM_2_DC_S,):` ? It is slightly less readable.

The way you wrote it only works because strings also have a weird list comprehension, as in
```python
>>> "a" in "aab"
True
>>> "aab" in "aab"
True
>>> "aabb" in "aab"
False
```
This is also why it works for line 215. If it was integers, this wouldn't work.

PR #1326, opened 2021-08-17
SW-261 Refactor model name logic and display S variant in about page.

amrsoll:
Screenshot from 2021-08-17 15-29-14

### DIFF
--- a/octoprint_mrbeam/util/device_info.py
+++ b/octoprint_mrbeam/util/device_info.py
@@ -68,7 +68,7 @@ class DeviceInfo(object):
     def get_model(self, refresh=False):
         """
         Gives you the device's model id like MRBEAM2 or MRBEAM2_DC
-        The value is soley read from device_info file (/etc/mrbeam)
+        The value is solely read from device_info file (/etc/mrbeam)
         and it's cached once read.
         :return: model
         :rtype: String
--- a/octoprint_mrbeam/util/device_info.py
+++ b/octoprint_mrbeam/util/device_info.py
@@ -18,6 +18,21 @@ def deviceInfo(use_dummy_values=False):
     return _instance
 
 
+MODEL_MRBEAM_2 = "MRBEAM2"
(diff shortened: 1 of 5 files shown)


octoprint_mrbeam/util/device_info.py
```diff
@@ -54,8 +65,20 @@ def get_serial(self):
     def get_hostname(self):
         return self._device_data.get(self.KEY_HOSTNAME)
 
-    def get_model(self):
-        return self._device_data.get(self.KEY_MODEL, self.MODEL_MRBEAM_2)
+    def get_model(self, refresh=False):
+        """
+        Gives you the device's model id like MRBEAM2 or MRBEAM2_DC
+        The value is soley read from device_info file (/etc/mrbeam)
```
khaledsherkawi:
```suggestion
        The value is solely read from device_info file (/etc/mrbeam)
```

octoprint_mrbeam/util/device_info.py
```diff
@@ -54,8 +65,20 @@ def get_serial(self):
...
+        and it's cached once read.
+        :return: model
+        :rtype: String
+        """
+        if refresh or self._model is None:
+            self._model = self._device_data.get(self.KEY_MODEL, MODEL_DEFAULT)
+        return self._model
+
+    def is_mrbeam2_dreamcut(self):
```
khaledsherkawi, line 80: I would rather have a different def for MrBeam DC S so that we can check for either model. I believe we're going to use that in other tickets

khaledsherkawi (changes requested):

octoprint_mrbeam/util/device_info.py
amrsoll: I only kept this function (it was defined previously) as a backwards compatibility. I don't think it is a good idea to keep those functions around.. Prefer using `MrBeamPlugin.get_model() == device_info.MODEL_MRBEAM_2_DC_S` if need be?

khaledsherkawi (approved):

PR #1332, opened 2021-08-20
Feature/sw 341 mr beam laser head s power parameters calibration

JFSolution:
fixed correction_factor for Laserhead model S
added laserhead_model and target_power to analytics laserhead_info

### DIFF
--- a/octoprint_mrbeam/printing/comm_acc2.py
+++ b/octoprint_mrbeam/printing/comm_acc2.py
@@ -2088,23 +2088,24 @@ class MachineCom(object):
             # Limit GCode input (in case users enter a too high value in the gcode)
             self._current_intensity = parsed_intensity
             if self._current_intensity > intensity_limit:
-                self._logger.warn(
+                self._logger.debug(
                     "gcode intensity higher as allowed max, will limit to max value - %s => %s",
                     self._current_intensity,
                     intensity_limit,
                 )
-                self._current_intensity = intensity_limit
+            self._current_intensity = min(intensity_limit, self._current_intensity)
 
             # Apply power correction factor and limit again (in case there is something wrong with the calculation of
             # the correction factor)
             if self._power_correction_factor > max_correction_factor:
-                self._logger.warn(
+                self._logger.debug(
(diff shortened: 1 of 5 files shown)


octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -2079,29 +2079,50 @@ def _replace_intensity(self, cmd):
...
 
             # Apply power correction factor and limit again (in case there is something wrong with the calculation of
             # the correction factor)
+            if self._power_correction_factor > max_correction_factor:
+                self._logger.warn(
+                    "Power correction factor higher as allowed max, will limit to max value - %s => %s",
+                    self._power_correction_factor,
+                    max_correction_factor,
+                )
+                self._power_correction_factor = max_correction_factor
```
amrsoll: Do we need a warning for _every single_ gcode command that goes over that max? I would recommend just to do

```python
self._power_correction_factor = min(self._power_correction_factor, max_correction_factor)
```

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -2079,29 +2079,50 @@ def _replace_intensity(self, cmd):
...
             parsed_intensity = int(intensity_cmd[1:])
 
             # Limit GCode input (in case users enter a too high value in the gcode)
             self._current_intensity = parsed_intensity
             if self._current_intensity > intensity_limit:
+                self._logger.warn(
+                    "gcode intensity higher as allowed max, will limit to max value - %s => %s",
+                    self._current_intensity,
+                    intensity_limit,
+                )
```
amrsoll: Same here, please do not log stuff for every single gcode command processed. This part of the code needs to be more lean. Refrain from doing IO with the SD card if you could.

amrsoll (changes requested):

octoprint_mrbeam/printing/comm_acc2.py
JFSolution: If we don't log it we won't see why it is not the expected value, how about I change it to a debug level, so if we investigate it we could see what happens

amrsoll (approved):

PR #1334, opened 2021-08-20
SW-338 Add laser model to custom materials

khaledsherkawi:


amrsoll (approved):

amrsoll:
the renaming of `device_model` doesn't interfere with saving custom materials for your device? I guess it doesn't get filtered... :thinking: 

Please make sure that custom materials that were saved previously are also compatible. If the fields need to be changed for those as well, I would recommend doing it over a migration.

PR #1336, opened 2021-08-24
SW-342 default material settings binding to laser head model - 2nd attempt

khaledsherkawi:


octoprint_mrbeam/util/material_csv_parser.py
```diff
@@ -3,6 +3,7 @@
 
 MRBEAM = "Mr Beam II"
 MRB_DREAMCUT = "MrB II Dreamcut"
+MRB_DREAMCUT_S = "MrB II Dreamcut S"
 MRB_READY = "MrB II Dreamcut Ready"  # not used yet
 MRB_DREAMCUT_NOT_VALIDATED = "Dreamcut (not validated)"
```
amrsoll: Actually you might need to add `"Dreamcut S (not validated)"` to the invalid keys... but perhaps it already filters out all of the strings containing `"not validated"` ... Do you know?

octoprint_mrbeam/util/material_csv_parser.py
```diff
@@ -93,7 +96,7 @@ def parse_csv(path=None, laserhead=MRBEAM):
             if colorcode:
                 current_color = colorcode
 
-            if not mrbeamversion in [MRBEAM, MRB_DREAMCUT, MRB_READY]:
+            if not mrbeamversion in [MRBEAM, MRB_DREAMCUT, MRB_DREAMCUT_S, MRB_READY]:
```
amrsoll, line 122: Ah yes, I see that it only considers lines with the correct `mrbeamversion`. I wonder if we could already filter out all of the unrelated mrbeam models here? It's a minor optimisation anyway.

octoprint_mrbeam/util/material_csv_parser.py
```diff
@@ -219,12 +222,15 @@ def parse_csv(path=None, laserhead=MRBEAM):
...
+    device_model_name = model_id_to_csv_name(device_model)
+    lh_to_device_mapping = ""
+    if device_model_name and device_model_name is MRB_DREAMCUT and laserhead_model == "S":
+        lh_to_device_mapping = MRB_DREAMCUT_S
+    elif device_model_name:
+        lh_to_device_mapping = device_model_name
+    if lh_to_device_mapping not in dictionary:
+        lh_to_device_mapping = DEFAULT_LASER
+    res = dict(materials=dictionary.get(lh_to_device_mapping, {}), laser_source=laserhead_model)
     return res
```
amrsoll: I think this might be simplified by having tuple-keys in the `convert` dictionary:

```py
convert = {
  (mrbeam_model, laserhead) : spreadsheet_name
}
```

This way you don't need the extra mapping of laserhead to mrbeam_model which gets a bit confusing to read.

It is also more readable and shows which are the "permitted" combinations of device models and laserheads.
Aaand you can also have `("MRBEAM2_DC_S", DEFAULT_LASER)` as a possibility.

amrsoll (dismissed):
Could be improved but otherwise good :+1:

octoprint_mrbeam/util/material_csv_parser.py
khaledsherkawi: It is already filtered out but for the matter of consistency and to remember it exists in the csv file I'll add it here

octoprint_mrbeam/util/material_csv_parser.py
khaledsherkawi: Code has been refactored

octoprint_mrbeam/util/material_csv_parser.py
khaledsherkawi: will not consider this optimization for now

octoprint_mrbeam/util/material_csv_parser.py
```diff
@@ -3,21 +3,46 @@
...
+            LASER_MODEL_S,
+        ): MRB_DREAMCUT_S,
+        (
+            octoprint_mrbeam.util.device_info.MODEL_MRBEAM_2_DC_R1,
+            DEFAULT_LASER_MODEL,
+        ): MRBEAM,
+        (
+            octoprint_mrbeam.util.device_info.MODEL_MRBEAM_2_DC_R2,
+            DEFAULT_LASER_MODEL,
+        ): MRBEAM,
```
amrsoll:
```py
from __future__ import absolute_imports

from . import device_info

# now use device_info.MODEL_MRBEAM2_DC_S
```

You can also disable the retarded black formatting with 
```py
# fmt: off

# write some unformatted code

# fmt: on
```

Just small bits :)

amrsoll (dismissed):
Nice, compact

PR #1337, opened 2021-08-28
SW-352 Incompatible custom material settings to be marked

khaledsherkawi:


octoprint_mrbeam/static/js/convert.js
```diff
@@ -616,25 +616,56 @@ $(function () {
...
+                if (customMaterials.hasOwnProperty(materialKey)) {
+                    let m = customMaterials[materialKey];
+                    if (('laser_model' in m) && !customMaterialsLaserModels.includes(m.laser_model)) {
+                        customMaterialsLaserModels.push(m.laser_model);
+                    } else if (!('laser_model' in m) &&
+                        ('laser_type' in m) &&
+                        m.laser_type === 'MrBeamII-1.0' &&
+                        !customMaterialsLaserModels.includes('0')) {
+                        customMaterialsLaserModels.push('0');
+                    }
```
amrsoll: I believe that - instead of supporting the legacy materials which don't have the `laser_model` key, we should be doing a migration of the materials in the backend (especially since it would be backwards compatible). Then we can remove this extra `else` if the key is not present in the settings.

If the key is not found, I would prefer to show a warning that it is from an unknown laserhead.

N.B. Keep in mind that the import function for the custom materials should also do this conversion - or not as we are not sure if it was from a MrB II or a DC device.

amrsoll (changes requested):
I believe a conversion should be added to the migration and possibly the material import function.

I know the change is slightly out of scope and would have been preferable in #1336 but that one was merged. It could also start a conversation about what to do with custom material backups.

octoprint_mrbeam/static/js/convert.js
```diff
@@ -616,25 +616,56 @@ $(function () {
...
+                }
+            }
+            return customMaterialsLaserModels.length > 1 ||
+                (customMaterialsLaserModels.length === 1 &&
+                    customMaterialsLaserModels[0] !== MRBEAM_LASER_HEAD_MODEL);
```
amrsoll: I don't know how you'd do it in JS, but I feel like this is more transparent / simple ? (python flavour)
```python
return any( laser != MRBEAM_LASER_HEAD_MODEL for laser in customMaterialsLaserModels )
```

octoprint_mrbeam/static/js/convert.js
```diff
@@ -616,25 +616,56 @@ $(function () {
...
                 if (m !== null) {
                     //					m.name = materialKey; // TODO i18n
                     if (m.name.toLowerCase().indexOf(q) >= 0) {
                         m.key = materialKey;
                         m.custom = true;
                         out.push(m);
                     }
+                    if (materialCompatibilityDisplay) {
+                        m.compatible = m.laser_model === MRBEAM_LASER_HEAD_MODEL ||
+                            (!('laser_model' in m) && MRBEAM_LASER_HEAD_MODEL === '0');
```
amrsoll: Please sanitize the material lists once instead of handling legacy cases all the time (see previous comments)

octoprint_mrbeam/materials.py
```diff
@@ -38,7 +39,7 @@ def get_custom_materials(self):
 
     def put_custom_material(self, key, material):
```
amrsoll, line 40: So this is also used when loading older custom materials from a file?

khaledsherkawi: No this is when you add new materials, import JSON file to restore materials, and when the migration script sanitize the already existing materias

amrsoll (approved):

PR #1340, opened 2021-09-01
SW-351 guide the user when choosing an incompatible custom material settings

khaledsherkawi:


octoprint_mrbeam/migrate.py
```diff
@@ -944,3 +955,24 @@ def hostname_helper_scripts(self):
...
+        it sets the value according to the latest laserhead
+        model updates
+        It also replaces 'model' key with 'device_model'
+        """
+        self._logger.info("start update_custom_material_settings")
+        custom_materials = materials(self.plugin).get_custom_materials()
+        for material_id, material_settings in custom_materials.items():
+            if (
+                "laser_type" in material_settings
+                and material_settings["laser_type"] == "MrBeamII-1.0"
```
amrsoll:
```py
if dict_get(material_settings, ['laser_type']) == "MrBeamII-1.0":
```

octoprint_mrbeam/migrate.py
```diff
@@ -944,3 +955,24 @@ def hostname_helper_scripts(self):
...
+            ):
+                material_settings["laser_type"] = '0'
+                material_settings["laser_model"] = material_settings.pop("laser_type")
```
amrsoll: More straight-forward
```py
material_settings["laser_model"] = '0'
del material_settings["laser_type"]
```

octoprint_mrbeam/migrate.py
```diff
@@ -944,3 +955,24 @@ def hostname_helper_scripts(self):
...
+            if "model" in material_settings:
+                material_settings["device_model"] = material_settings.pop("model")
+            materials(self).put_custom_material(material_id, material_settings)
```
amrsoll: wrong argument. Doesn't matter here because it is pointing to a filthy global variable.

If I were you:
```py
my_materials = materials(self.plugin)
for material_id, material_settings in my_materials.get_custom_materials().items():
    ...
    my_materials.put_custom_material(material_id, material_settings)

File: octoprint_mrbeam/static/js/convert.js [hunk 4]
```diff
@@ -146,6 +147,13 @@ $(function () {
             var mat = self.selected_material();
             if (mat !== null) return mat === null ? "" : mat.img;
         });
+        self.selected_material_compatibility = ko.computed(function () {
```
amrsoll, line 149: Variable name declared 2x (as an observable and as a computed function...)

octoprint_mrbeam/static/js/convert.js
```diff
@@ -146,6 +147,13 @@ $(function () {
...
+            if (self.selected_material() !== null &&
+                'compatible' in self.selected_material() &&
+                !self.selected_material().compatible) {
+                return false;
+            }
```
amrsoll: What about the `else` ? Is the function just supposed to return `false` or `null` ?

octoprint_mrbeam/static/js/convert.js
```diff
@@ -182,9 +190,18 @@ $(function () {
...
                 } else {
-                    self.save_custom_material_name(
-                        self.selected_material().name
-                    );
+                    if ('compatible' in self.selected_material() &&
+                        !self.selected_material().compatible) {
+                        self.save_custom_material_name(
+                            gettext("Laser Model") + " "
+                            + MRBEAM_LASER_HEAD_MODEL + " - " + self.selected_material().name
+                        );
```
amrsoll: So .. couldn't we change the default meaning of `compatible` if it is not present? It would be `null` right?
So what if `!selected_material().compatible` is only `true` if the material is _explicitely_ compatible? I could understand if the default materials do not have this key, but I think all of them should just be labeled compatible explicitely.

octoprint_mrbeam/static/js/convert.js
```diff
@@ -616,25 +634,53 @@ $(function () {
             self._update_job_summary();
         });
 
+        self.showMaterialCompatibility = function () {
+            let customMaterialsLaserModels = [];
+            let customMaterials = self.custom_materials();
+            // Get all laser models used while saving custom materials
+            for (let materialKey in customMaterials) {
+                if (customMaterials.hasOwnProperty(materialKey)) {
```
amrsoll: isn't this redundant? If the `materialKey` is `in customMaterials`, then `customMaterials.hasOwnProperty(materialKey)` would always be true?

octoprint_mrbeam/static/js/convert.js
```diff
@@ -616,25 +634,53 @@ $(function () {
...
+                    let m = customMaterials[materialKey];
+                    if (('laser_model' in m) && !customMaterialsLaserModels.includes(m.laser_model)) {
+                        customMaterialsLaserModels.push(m.laser_model);
+                    }
```
amrsoll: You have already sanitised the materials, you shouldn't have to check if `laser_model` is an attribute anymore.

octoprint_mrbeam/static/js/convert.js
```diff
@@ -616,25 +634,53 @@ $(function () {
...
+                }
+            }
+            return customMaterialsLaserModels.some(item => item !== MRBEAM_LASER_HEAD_MODEL);
```
amrsoll: So in conclusion, wouldn't this function just return 
```js
return customMaterials.some(item => item.laser_model !== MRBEAM_LASER_HEAD_MODEL);
```
?

octoprint_mrbeam/static/js/convert.js
```diff
@@ -616,25 +634,53 @@ $(function () {
...
                     if (m.name.toLowerCase().indexOf(q) >= 0) {
                         m.key = materialKey;
                         m.custom = true;
                         out.push(m);
                     }
+                    if (materialCompatibilityDisplay) {
+                        m.compatible = m.laser_model === MRBEAM_LASER_HEAD_MODEL;
+                        if (m.laser_model === 'S') {
+                            m.customBeforeElementContent = '[S]';
+                        }
```
amrsoll: Should the `[S]` symbol _only_ show when other material settings are incompatible, or all the time?

From the Jira ticket, I feel like it is up for interpretation.. there is no guideline for what to show when there is no compatibility problem.

amrsoll: And if we want to show the `[S]` symbol all the time, then we don't need to check the complete `materialCompatibilityDisplay` here.

octoprint_mrbeam/static/js/settings/custom_material.js
```diff
@@ -20,6 +20,19 @@ $(function () {
                 self.readLocalFile(ev, function (data) {
                     if (data && "custom_materials" in data) {
                         console.log("Loaded material settings: ", data);
+                        //  Update custom material settings exported with new keys and laser model value
+                        for (let [material_id, material_settings] of Object.entries(data.custom_materials)) {
+                            if ('laser_type' in material_settings &&
+                                material_settings['laser_type'] === "MrBeamII-1.0") {
+                                material_settings['laser_type'] = '0';
+                                material_settings['laser_model'] = material_settings['laser_type'];
+                                delete material_settings['laser_type'];
```
amrsoll: See previous comment. No need to re-write a value if you are going to delete it.

Also - I had hoped that the sanitisation function would be the same for the front and backend (By same I mean that it is the same piece of code, not a mirror of one-another)

amrsoll (changes requested):

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: yea but this is an IDE suggestion the was I just accepted which doesn't make sense here. I removed it.

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: I would still leave this as an extra check in case we missed something. This way no error will be thrown

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: yea the whole function isn't needed anymore. This has been removed.

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: No this has been clarified and added in the design ticket. The [S] symbol will be showing only when the S model is detected in the custom materials. It can be compatible or not

amrsoll: Disagree. It is not explicitely stated - it is only written that these icons should show in case of a clash; but that doesn't invalidate my question. In any case it seems to be fine :)

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: fixed

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: fixed via the previous feedback

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: I preferred to leave it out whenever the whole compatibility issue isn't existing

octoprint_mrbeam/templates/conversion_dialog.jinja2
```diff
@@ -193,7 +193,7 @@
                     <p>
                         <i class="icon-warning-sign" style="color:red;"
                            data-bind="visible: material_safety_notes()"></i>
-                        <span id="safety_notes" class="" data-bind="text: material_safety_notes"></span>
+                        <span id="safety_notes" class="" data-bind="text: material_safety_notes, css:{ incompatible_safety_warning: $root.selected_material_compatibility() === false }"></span>
```
amrsoll: personaly prefer `!something` rather then `something === false` - but perhaps it is because of `null` behaviour?

khaledsherkawi: resolved

amrsoll (approved):

PR #1342, opened 2021-09-03
SW-350 one time show a guide to the user if laser head s model is detected

khaledsherkawi:


octoprint_mrbeam/static/js/laserhead_changed.js
```diff
@@ -0,0 +1,60 @@
...
+    function LaserheadChangedViewModel(params) {
+        let self = this;
+        window.mrbeam.viewModels["LaserheadChangedViewModel"] = self;
+
+        self.settings = params[0];
+        self.loginState = params[1];
+
+        self.onUserLoggedIn = function () {
+            if (
+                self.loginState.currentUser?.()?.active
```
amrsoll, line 11: looks interesting, could `currentUser` not exist? Do you need the extra `?.` ?

octoprint_mrbeam/static/js/laserhead_changed.js
```diff
@@ -0,0 +1,60 @@
...
+            ) {
+                let laserheadChanged = self.settings.settings.plugins.mrbeam.laserheadChanged();
+                if (laserheadChanged) {
```
amrsoll: Seems like it can be shortened

octoprint_mrbeam/static/js/laserhead_changed.js
```diff
@@ -0,0 +1,60 @@
...
+                                "Unable to save laserhead change detection at the moment.%(br)sCheck connection to Mr Beam and try again."
+                            ),
+                            { br: "<br/>" }
+                        ),
+                        type: "error",
+                        hide: true,
+                    });
+                });
+        }
+    }
```
amrsoll: What is the purpose of this? Why is the front end telling the back-end whether the laserhead got changed?

amrsoll: If the data is static... Couldn't we ignore it all together?

octoprint_mrbeam/static/js/laserhead_changed.js
```diff
@@ -0,0 +1,60 @@
+$(function () {
+    function LaserheadChangedViewModel(params) {
```
amrsoll, line 2: I think you will need to argument for why there is a separate viewmodel for this feature.

octoprint_mrbeam/static/js/laserhead_changed.js
amrsoll: I think I understand what this is about now... Could you rename it to `laserheadChangeAcknowledged()` ?

octoprint_mrbeam/__init__.py
```diff
@@ -1918,6 +1926,8 @@ def on_api_command(self, command, data):
             return NO_CONTENT
         elif command == "focus_reminder":
             return self.focus_reminder(data)
+        elif command == "laserhead_changed":
+            return self.laserhead_changed(data)
```
amrsoll: ... so, in the backend, you can receive the user acknowledgment, and then decide to 
```py
self._settings.set_boolean(["laserheadChanged"], dict_get(data, "laserhead_change_acknowledged", False)
```
I do not see the need for a separate `laserhead_changed()` function.

amrsoll (changes requested):

octoprint_mrbeam/static/js/laserhead_changed.js
khaledsherkawi: Yes I think that happens in the scenario when the session expires

octoprint_mrbeam/static/js/laserhead_changed.js
khaledsherkawi: done

octoprint_mrbeam/static/js/laserhead_changed.js
khaledsherkawi: I was think about Acknowledged as well ^^ 
Done

octoprint_mrbeam/static/js/laserhead_changed.js
khaledsherkawi: Well that's the way we are handling this for now (check focus reminder). But I agree we need to revamp this at some point and have all such windows belong to one View model.

octoprint_mrbeam/__init__.py
khaledsherkawi: I'm not sure what you mean here

amrsoll: basically it was shortening it too

amrsoll: nvm

octoprint_mrbeam/static/js/laserhead_changed.js
amrsoll: Could you also reflect the new name in the json you send please?

amrsoll: Or do not send any "data", just send back empty data with the API call `laserhead_change_acknowledged`

khaledsherkawi: No I think that stands correct. The config parameter reflect an event happening which is laser being changed

amrsoll: And what is that for?
`self.settings.requestData();`

amrsoll: Yes, in the config itself, not in the front-end. The front end only knows that the user acknowledged the change, but it is up to the backend to choose whether to do something with the acknowledgement

khaledsherkawi: self.settings.requestData() has been removed

khaledsherkawi: I agree with you, I just don't see what the need to change the parameter name into acknowledged in the backend

amrsoll (approved):

PR #1344, opened 2021-09-07
SW-420 final feedback on the frontend ux

khaledsherkawi:


octoprint_mrbeam/static/js/convert.js
```diff
@@ -252,6 +246,9 @@ $(function () {
         self.save_material_settings = function () {
             var name = self.save_custom_material_name();
             var key = self._replace_non_ascii(name).toLowerCase();
+            if(self.incompatible_custom_materials_keys().includes(key)){
+                key = self._replace_non_ascii(MRBEAM_LASER_HEAD_SERIAL).toLowerCase() + " " + key;
+            }
```
amrsoll: So if I make a custom material on top of a custom material etc.... will it be saved as 
`laserhead S laserhead S laserhead S my custom material` ?

octoprint_mrbeam/static/js/convert.js
```diff
@@ -308,6 +305,9 @@ $(function () {
                 new_material.device_model = MRBEAM_MODEL;
                 new_material.custom = true;
                 new_material.v = BEAMOS_VERSION;
+                // Remove compatibility keys
+                delete new_material.compatible;
+                delete new_material.customBeforeElementContent;
```
amrsoll, line 317: Why? I could understand if you would set the keys to be `new_material.compatible = true`. If you don't need the keys, then why have them in the 1st place?

I guess `some_material.compatible = ko.computed(function() { return MRBEAM_LASER_HEAD_MODEL == some_material.laser_head }` ?

octoprint_mrbeam/static/js/convert.js
```diff
@@ -370,7 +370,7 @@ $(function () {
                     var fm = self.filteredMaterials();
                     for (var i = 0; i < fm.length; i++) {
                         var my_material = fm[i];
-                        if (my_material.name === new_material.name) {
+                        if (my_material.key === new_material.key) {
```
amrsoll: This is not the reference key in the material mapping. I just discovered that the `key` attribute exists for the material settings but I have no idea what it is for.

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: No, this is a bug I found where upon saving custom materials with the compatibility keys available then these will get saved within the yaml file. So here I'm just preventing that from happening because it would lead to a bug whenever you switch laserheads again and go back to older laserhead.

amrsoll: You should probably sanitise that when the custom material gets saved in the back end

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: So first thing we prepend via the serial number and not the model. Secondly, this only happens when you save a material with an incompatible material name. Upon saving again with the same name, it won't get prepended again but rather it would add the new custom materials to the already saved key: serial + name.

There is a corner case where this can happen if the user flipped the laserheads again and tried to save the incompatible custom settings while the compatible ones has the same name then this will happen and a new key with the SERIAL + SERIAL + name will be saved => This was just resolved

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: This turned out to be faulty change. It's fixed now.

octoprint_mrbeam/static/js/convert.js
khaledsherkawi: I would leave it in the frontend as the whole concept of compatibility exist there

amrsoll: Please make sure that the owner of the data (in this case the back-end) makes a sanity check on the data it is given. In this case I would filter away any key we do not expect. We could possibly be doing a more extensive check on the types etc but that is not quite required right now

amrsoll: in this case it could be
```python
accepted keys = ['name', 'key', ...]
result_sanitised = {}
for mat_key, mat_info in json_from_api_call.items():
  result_sanitised[mat_key] = {k: v for k,v in mat_info.items() if k in accepted_keys}
```

amrsoll: N.B. This would also prevent your compatibility bug when changing laserheads.

khaledsherkawi: Done

amrsoll (dismissed):

amrsoll (approved):

PR #1352, opened 2021-09-13
SW-176 update of grbl to fix burnmarks issue

JFSolution:


octoprint_mrbeam/migrate.py
```diff
@@ -31,7 +31,7 @@ class Migration(object):
     VERSION_PREFILL_MRB_HW_INFO = "0.1.55"
     VERSION_AVRDUDE_AUTORESET_SCRIPT = "0.2.0"
     VERSION_USERNAME_LOWCASE = "0.2.0"
-    VERSION_GRBL_AUTO_UPDATE = "0.2.1"
+    VERSION_GRBL_AUTO_UPDATE = "0.9.9"
```
ahmed-mrbeam: I believe we agreed that this change would be released with 0.10.0 alpha

khaledsherkawi: This will still work as the comparison function "equal_ok" is set to False so this will be triggered on any version higher than 0.9.9

khaledsherkawi (dismissed):

octoprint_mrbeam/migrate.py
ahmed-mrbeam: higher or equal to 0.9.9, right ?
if so, then it will be misleading as we will not release this with 0.9.9 which means it is not tested properly with 0.9.9

khaledsherkawi: It can be misleading and it could be misread.

(VERSION_GRBL_AUTO_UPDATE = "0.10.0"
with the parameter "equal_ok" set to True)
=> can be more readable

JFSolution: No, the equal_ok is needed if the migration should be done if the version previous and the needed version is the same
So if I set the `VERSION_GRBL_AUTO_UPDATE` to 0.10.0 the migration will be done for all mrbeam plugins, that will update from all version below 0.10.0 but will not be done if the version before the migration is 0.10.0.

to make it more clear here the migrations between different versions:
**0.9.8 -> 0.10.0**
```
 12:54:26 octoprint.plugins.mrbeam.migrate INFO Starting migration from v0.9.8 to v0.10.0
 12:54:26 octoprint.plugins.mrbeam.migrate INFO auto_update_grbl()
 12:54:26 mrbeam.printing.profile WARNING Not selecting the _default profile because of OP default behaviour. See ``octoprint_mrbeam.printing.profile.select()``.
 12:54:26 octoprint.plugins.mrbeam.migrate INFO start update_custom_material_settings
 12:54:26 octoprint.plugins.mrbeam.migrate INFO Finished migration from v0.9.8 to v0.10.0.
```

**0.9.9 -> 0.10.0**
```
  13:54:57 octoprint.plugins.mrbeam.migrate INFO Starting migration from v0.9.9 to v0.10.0
 13:54:57 octoprint.plugins.mrbeam.migrate INFO auto_update_grbl()
 13:54:57 mrbeam.printing.profile WARNING Not selecting the _default profile because of OP default behaviour. See ``octoprint_mrbeam.printing.profile.select()``.
 13:54:57 octoprint.plugins.mrbeam.migrate INFO Finished migration from v0.9.9 to v0.10.0.
```

**0.10.0 -> 0.10.0**
```
 14:00:44 octoprint.plugins.mrbeam.migrate DEBUG beomosdate 2021-07-22 version 0.10.0
 14:00:44 octoprint.plugins.mrbeam.migrate DEBUG No migration required.
```

**0.10.0 -> 0.10.1**
```
14:02:20 octoprint.plugins.mrbeam.migrate INFO Starting migration from v0.10.0 to v0.10.1
 14:02:20 octoprint.plugins.mrbeam.migrate INFO Finished migration from v0.10.0 to v0.10.1.
```

**0.10.0 -> 0.9.7**
```
14:03:44 octoprint.plugins.mrbeam.migrate DEBUG No migration required.
```

khaledsherkawi: Yes you are right. Sounds good.

khaledsherkawi (approved):

PR #1356, opened 2021-09-15
SW-176 add grbl latest version to mr beam plugin

JFSolution:
update of https://github.com/mrbeam/MrBeamPlugin/pull/1352

this will fix that new images won't update grbl


khaledsherkawi (dismissed):

khaledsherkawi:
@Josef-MrBeam I asked Axel to have a quick look as well. So please don't merge without his approval.

octoprint_mrbeam/migrate.py
```diff
@@ -571,10 +571,25 @@ def update_mount_manager(
...
+            self.plugin._settings.set(
+                ["grbl", "version_lastknown"],
+                self.plugin._settings.get(["grbl_version_lastknown"]),
+                force=True,
+            )
+            self.plugin._settings.set(
+                ["grbl_version_lastknown"],
+                None,
+                force=True,
+            )
```
amrsoll: use `settings.remove("grbl_version_lastknown")`

octoprint_mrbeam/migrate.py
```diff
@@ -571,10 +571,25 @@ def update_mount_manager(
...
+        self.plugin._settings.set(
+            ["grbl", "auto_update_version"], self.GRBL_AUTO_UPDATE_VERSION, force=True
+        )
+        self.plugin._settings.set(
+            ["grbl", "auto_update_file"], self.GRBL_AUTO_UPDATE_FILE, force=True
+        )
+        self.plugin._settings.save()
```
amrsoll: you are using `self.plugin._settings` a lot. I would recommend to assign it to a variable for this code here, like `settings = self.plugin._settings`

This can also help fit the lines inside the 80 character limit.

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -336,18 +336,31 @@ def _monitor_loop(self):
...
+                    _mrbeam_plugin_implementation._settings.get(
+                        ["grbl", "auto_update_file"]
+                    ),
                 )
                 self.flash_grbl(
-                    grbl_file=self._laserCutterProfile["grbl"]["auto_update_file"],
+                    grbl_file=_mrbeam_plugin_implementation._settings.get(
+                        ["grbl", "auto_update_file"]
+                    ),
                     is_connected=False,
```
amrsoll: Same here

`settings = self._mrbeam_plugin_implementation._settings`

Although accessing the global variable `_mrbeam_plugin_implementation` is Not recommended.
To fix that I recommend to move this function out of the `comm_acc2` module. Responsability-wise, this is up to the `Laser` object to ask the `Laser.disconnect()`, which closes the `_comm` object, then run that shell command. After that you can try to re-`connect()` from the `Laser` class.

The problem is that the `Laser` class only has access to the "printer profile" or - in our case - "laser profile". And this makes sense in case we have the same general OctoPrint settings in the plugin, but we want to adapt it to different machines, then this config is the place where we isolate the changes.

Hopefully you can manage to migrate this responsability to the `Laser` class.

Please keep the grbl settings in the laser cutter profile. This can require quite some migration such as `grbl_version_lastknown` into that profile too.

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1894,37 +1907,45 @@ def _get_grbl_file_name(grbl_version=None):
...
                     )
+                    _mrbeam_plugin_implementation._settings.set(
+                        ["grbl", "auto_update_file"], None, force=True
+                    )
+                    _mrbeam_plugin_implementation._settings.save()
                 except:
                     self._logger.exception(
-                        "Exception while saving lasercutterProfile changes to auto update controls: "
+                        "Exception while saving Mr Beam settings changes for auto update controls"
                     )
```
amrsoll: Please just remove useless exceptions

amrsoll (changes requested):
After I FINALLY understood what went wrong (whoops :sweat_smile: ), I think I found the problem! The laser cutter profile could not load properly because `LaserCutterProfile.get(identifier)` could not find the `identifier` in the `LASER_PROFILE_IDENTIFIERS` - which don't include the latest identifiers, namely `"MrBeam2S"` and `"MrBeam2R"`. You can add those profiles in the `octoprint_mrbeam/printing/profiles/`.

After that , I thiiink the GRBL auto-updates should work. But I believe my comments on refactoring still apply.

amrsoll:
Actually, if you have a look at `octoprint.printer.profile.PrinterProfileManager.save()`, then you see that you can ask it to fire an Event. You can subscribe to that Event in the `Laser` class. If the grbl version of the current profile doesn't match the installed one, then you can start flashing the new grbl version.

This allows us to 
- remove all grbl stuff from the `__init__` file, 
- remove `reset_grbl_auto_update_config` because we compare versions to what we expect
- `auto_update_grbl()` only needs to know what the new version is. No need to compare to anything else etc...

octoprint_mrbeam/migrate.py
JFSolution: yes, good point

octoprint_mrbeam/migrate.py
JFSolution: this will only save 13 signs :), so I think it is not urgend, but yes I can do it :)

octoprint_mrbeam/printing/comm_acc2.py
JFSolution: I didn't add it, and I think there is a reason why it is there, so I just changed the message

JFSolution:
> After I FINALLY understood what went wrong (whoops 😅 ), I think I found the problem! The laser cutter profile could not load properly because `LaserCutterProfile.get(identifier)` could not find the `identifier` in the `LASER_PROFILE_IDENTIFIERS` - which don't include the latest identifiers, namely `"MrBeam2S"` and `"MrBeam2R"`. You can add those profiles in the `octoprint_mrbeam/printing/profiles/`.
> 
> After that , I thiiink the GRBL auto-updates should work. But I believe my comments on refactoring still apply.

Nop, that won't fix it, because we removed the profile file thing, it will work now completly without a profile file, it only uses the coded profile and it seems like there is no way to add anything there while it is running

octoprint_mrbeam/printing/comm_acc2.py
amrsoll: There are only bad reasons for having the exception exactly here. As a rule :
1. Only catch the exceptions you can foresee
2. If exceptions don't escalate or get logged (in the example of threads for example) - then you can have the luxury of just logging all the exceptions.

In every case where you catch every exception, you Should raise it again.

I know it is not your code originally, but when we touch some function, we want to leave it in a better state than we started with. Therefore we should also make sure to refactor bad practices.

amrsoll:
Yes and no. It will still use the files over the coded defaults. If you have a file with the correct ID, then it will overwrite the default settings.
That is why you can still save the config and it should override the hardcoded configs.

octoprint_mrbeam/__init__.py
```diff
@@ -2674,7 +2674,9 @@ def get_mrb_state(self):
                     dusting_mode=self.dust_manager.is_final_extraction_mode,
                     state=self._printer.get_state_string(),
                     is_homed=self._printer.is_homed(),
-                    laser_model = self.laserhead_handler.get_current_used_lh_data()["model"],
+                    laser_model=self.laserhead_handler.get_current_used_lh_data()[
```
ahmed-mrbeam, line 2677: is this format change only ?

ahmed-mrbeam:
@Josef-MrBeam  is there a reason why we deleted all these GRBL files ? is this only cleanup ?

octoprint_mrbeam/migrate.py
```diff
@@ -579,10 +579,14 @@ def update_mount_manager(
...
     def auto_update_grbl(self):
         self._logger.info("auto_update_grbl() ")
-        default_profile = laserCutterProfileManager().get_default()
-        default_profile["grbl"]["auto_update_version"] = self.GRBL_AUTO_UPDATE_VERSION
-        default_profile["grbl"]["auto_update_file"] = self.GRBL_AUTO_UPDATE_FILE
-        laserCutterProfileManager().save(default_profile, allow_overwrite=True)
+        # migrate version last_known
+        laserCutterProfile = laserCutterProfileManager().get_current_or_default()
+        self._logger.debug("profile: %s", laserCutterProfile)
+        laserCutterProfile["grbl"][
```
ahmed-mrbeam: what if laserCutterProfile is None ??

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1826,7 +1826,7 @@ def flash_grbl(self, grbl_file=None, verify_only=False, is_connected=True):
                         title="GRBL Update failed", text=msg, is_err=True
                     )
                 )
-            except:
+            except Exception:
```
ahmed-mrbeam, line 1829: That is very broad, are there specific set of exceptions we anticipate ?

octoprint_mrbeam/printing/profile.py
```diff
@@ -191,6 +192,11 @@ def get(self, identifier):
                 hard_coded = dict_merge(default, LASER_PROFILE_MAP[identifier])
                 return dict_merge(hard_coded, file_based_result)
             else:
+                if identifier is None:
```
ahmed-mrbeam, line 200: why this if statement is inside the else statement ?, why not on the same level as elif ?

octoprint_mrbeam/printing/profile.py
```diff
@@ -234,7 +240,8 @@ def get_current_or_default(self):
         return PrinterProfileManager.get_current_or_default(self)
 
     def exists(self, identifier):
-        if identifier in LASER_PROFILE_IDENTIFIERS:
+        if re.match(r"MrBeam+(\d){1}[A-Z]", identifier):
```
ahmed-mrbeam: more explanation in a code comment please, why we changed the old check completely and using regex now ?

octoprint_mrbeam/util/device_info.py
```diff
@@ -59,6 +59,9 @@ def get(self, key, default=None):
     def get_series(self):
         return self._device_data.get(self.KEY_DEVICE_SERIES)
 
+    def get_type(self):
```
ahmed-mrbeam, line 62: type is ?.... laser head model ?

JFSolution: its the device type for example `MrBeam2S`

octoprint_mrbeam/__init__.py
JFSolution: yes

octoprint_mrbeam/printing/profile.py
JFSolution: because it would use the same return, so I put them together so it wont't have a copy of it

octoprint_mrbeam/printing/profile.py
JFSolution: ok I can add this

octoprint_mrbeam/migrate.py
ahmed-mrbeam: @Josef-MrBeam  this one is still not answered

octoprint_mrbeam/printing/comm_acc2.py
ahmed-mrbeam: @Josef-MrBeam

JFSolution: As discusses today we have to figure out how we deal with such broad exceptions

ahmed-mrbeam (dismissed):

amrsoll:
All good except for 2 things:
- Please save the dictionary diff between the default and the changed config - This way we don't overwrite the default config if we decide to modify the default values, unless we specifically decide to save these default profile values in the yaml file
- Please check with @ahmed-mrbeam if it's ok to downgrade the version from 0.10.x to 0.9.x

octoprint_mrbeam/printing/profile.py
```diff
@@ -263,14 +268,51 @@ def _load_default(self, defaultModel=None):
             raise InvalidProfileError()
         return profile
 
+    def _save_to_path(self, path, profile, allow_overwrite=False):
```
amrsoll, line 271: Alright, now please add docstring + what changed compared to the overloaded function

amrsoll (approved):

PR #1357, opened 2021-09-16
SW-346 warn users not to remove laser head while the device is on - 2

khaledsherkawi:


amrsoll:
What was wrong? Didn't save it soon enough?

khaledsherkawi:
> What was wrong? Didn't save it soon enough?
@amrsoll 
There was an exception thrown because of the way we were using the "save()" method

amrsoll (approved):
Ah yeah of course, I should have caught that it was Js style :facepalm: my bad

PR #1349, opened 2021-09-10
Feature/sw 381 mr beam add laserhead model in analytics

JFSolution:
SW-381 added laser head model to analytics (device/startup, job/laserjob_started, job/conv_eng, job/conv_cut, job/dust_summary, job/intensity_summary, job/lasertemp_summary)

PR #1358, opened 2021-09-16
SW-467 fix 'S' for laserhead model in analytics data

JFSolution:
fix for https://github.com/mrbeam/MrBeamPlugin/pull/1349 (SW-381)


octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -38,8 +39,8 @@ def __init__(self, plugin):
         self._load_laser_heads_file()  # Loads correction_settings, last_used_lh_serial and lh_cache
 
         self._current_used_lh_serial = self._last_used_lh_serial
-        self._current_used_lh_model = self._last_used_lh_model
-        self._current_used_lh_model_id = None
+        self._current_used_lh_model_string = self._last_used_lh_model_string
+        self._current_used_lh_model_id = self._last_used_lh_model_id
```
amrsoll: That is nice, thank you for adding the consistency :)

amrsoll: However, if the model id and the model string are refering to the same resource, then the derivative values should be observing the source and shouldn't be variables themselves.

It would be more prudent that `self._current_used_lh_model_id` is the source variable and ` self._current_used_lh_model_string` becomes a function which uses the ` self._current_used_lh_model_id`. You can turn it into a `property` which allows it to be used the same way you would normally do with an object class property.

```python
@property
def  _current_used_lh_model_string(self):
    match self._current_used_lh_model_id:
        case 0: return MODEL_DC_STR
        case 1: return MODEL_S_STR
        case _: raise ValueError("Unknown laserhead model ID {!r}".format(self._current_used_lh_model_id))

Comment (amrsoll) [reply octoprint_mrbeam/iobeam/laserhead_handler.py]: Same goes for `_last_used_lh_model_string` - this way you never forget to update the origin resource and the derivative values :)

File: octoprint_mrbeam/iobeam/laserhead_handler.py [hunk 2]
```diff
@@ -251,7 +252,7 @@ def _calculate_power_correction_factor(self):
 
             # laserhead model S fix for correction factor
             # TODO fix this GOAL_MAX problem for all laser heads in a separate issue SW-394
-            if self._current_used_lh_model == "S":
+            if self._current_used_lh_model_string == "S":
```
amrsoll: Also Please use constants instead of magic values.

octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -325,8 +326,13 @@ def _load_laser_heads_file(self):
...
+                    if "last_used_lh_model_string" in data:
+                        self._last_used_lh_model_string = data[
+                            "last_used_lh_model_string"
+                        ]
+
+                    if "last_used_lh_model_id" in data:
+                        self._last_used_lh_model_id = data["last_used_lh_model_id"]
 
                     if "correction_settings" in data:
                         self._correction_settings = data["correction_settings"]
```
amrsoll: Please refactor to remove all of these if statements

amrsoll: And remove the general try / except as well. You may catch OSError or similar which can happen when the file is not present. You may also catch errors when the Yaml file has incorrect syntax.
And to prevent that error from happening, make sure to use `os.path.isfile`.

octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -342,7 +348,8 @@ def _write_laser_heads_file(self, file=None):
             laser_heads=self._lh_cache,
             correction_settings=self._correction_settings,
             last_used_lh_serial=self._current_used_lh_serial,
-            last_used_lh_model=self._current_used_lh_model,
+            last_used_lh_model_string=self._current_used_lh_model_string,
+            last_used_lh_model_id=self._current_used_lh_model_id,
```
amrsoll: Please do not write 2 values which could potentially clash with each other. If `_current_used_lh_model_string` is a derivative of `_current_used_lh_model_id` then it should not be saved. At most 1 of the 2 should be saved, but not both.

amrsoll (changes requested):

octoprint_mrbeam/iobeam/laserhead_handler.py
JFSolution: That is what I said in the standup and we talked about it and the result was to have both saved

amrsoll: ah.. what was the reason already?

octoprint_mrbeam/iobeam/laserhead_handler.py
amrsoll: Actually I would recommend to use a mapping instead of a match / case.

octoprint_mrbeam/iobeam/laserhead_handler.py
JFSolution: This is only there till SW-394 will fix this bug, so the constant won't be needed

amrsoll: Why would sw-394 (ticket for the power correction settings) fix this? Why not fix it now?

JFSolution: The power calibration is still not activated for the dreamcut devices, only for the [S] laser heads.
To fix it for the dreamcut laserheads we have to adjust the custom material settings or inform the customers that they have to check the settings.

octoprint_mrbeam/iobeam/laserhead_handler.py
JFSolution: So why I should remove it then? If errors can appera here we should catch them

amrsoll: yes, but only the errors you foresee to happen. catching all exceptions are an absolute no-no otherwise.

amrsoll:
Please rebase the branch onto the latest commit in develop to fix the diff preview.

octoprint_mrbeam/iobeam/laserhead_handler.py
amrsoll: you can use `@logExceptions` to log and re-raise the exceptions

amrsoll: As you can see the exceptions are caught everywhere...

This is the typical example of every single function catching their own exceptions. You catch them at the lowest level where the stack trace is too small to get any context. Not only that but a lot of the time we don't even get a stacktrace...

It doesn't make sense to catch all exceptions ALL THE TIME, period.

```python
    def _handle_laserhead(self, dataset):
        """
        Handle laserhead dataset, iobeam sends the whole laserhead data.
        :param dataset:
        :return: error count
        """
        try:
            self._laserhead_handler.set_current_used_lh_data(dataset)
        except:
            self._logger.exception("laserhead: exception while handling head:data: ")
        return 0
```

```python
    def _handle_dataset(self, name, dataset):
        """
        Handle dataset
        :param name: name of the dataset
        :param dataset: the contents of the dataset
        :return: error count
        """
        # self._logger.info("ANDYTEST _handle_dataset() %s: %s", name, dataset)
        error_count = 0
        processing_start = time.time()

        err = -1
        try:
            if len(name) <= 0:
                err = self._handle_invalid_dataset(name, dataset)
                # [ ... ]
            else:
                if name == self.DATASET_FAN_DYNAMIC:
                    err = self._handle_fan_dynamic(dataset)
                # [ ... ]
                elif name == self.DATASET_LASERHEAD:
                    err = self._handle_laserhead(dataset)
                # [ ... ]
                else:
                    err = self._handle_unknown_dataset(name, dataset)
        except:
            self._logger.exception("Error handling dataset '%s': %s", name, dataset)

        if err >= 0:
            error_count += err

        processing_time = time.time() - processing_start
        self._handle_processing_time(processing_time, dataset, err)

        return error_count
```

```python
    def _handle_messages(self, messages):
        """
        Handles incoming list of messages from the socket.
        :param messages: list of incoming dict messages
        :return: int: number of invalid messages 0 means all messages were handled correctly
        """
        error_count = 0
        message_count = 0
        try:
            for json_data in messages:
                if len(json_data) > 0:

                    message_count = +1

                    try:
                        json_dict = json.loads(json_data)
                        # Now there could be "data" and "response"
                        if "data" in json_dict:
                            _data = json_dict["data"]
                            if self.MESSAGE_ERROR not in _data:
                                # Process all data sets
                                if isinstance(_data, dict):
                                    # We have to process the iobeam dataset first, because we need the iobeam version for analytics
                                    if "iobeam" in _data.keys():
                                        error_count += self._handle_dataset(
                                            "iobeam", _data.pop("iobeam", None)
                                        )

                                    for data_id, dataset in _data.items():
                                        error_count += self._handle_dataset(
                                            data_id, dataset
                                        )
                            else:
                                self._logger.debug(
                                    "Received error in data '%s'",
                                    _data[self.MESSAGE_ERROR],
                                )
                                error_count += 1
                        elif "response" in json_dict:
                            error_count += self._handle_response(json_dict)

                    except ValueError as ve:
                        # Check if we communicate with an older version of iobeam, iobeam:version:0.6.0
                        if isinstance(json_data, basestring) and json_data.startswith(
                            "iobeam:version:"
                        ):
                            tokens = json_data.split(":")
                            try:
                                self.iobeam_version = tokens[2]
                                version_obj = LooseVersion(tokens[2])
                            except ValueError:
                                self.iobeam_version = None
                                version_obj = None
                                self._logger.debug(
                                    "Could not parse iobeam version and data '%s' as JSON",
                                    json_data,
                                )

                        else:
                            self._logger.debug(
                                "Could not parse data '%s' as JSON - err : %s",
                                json_data,
                                ve,
                            )
                    except Exception as e2:
                        self._logger.debug("Some error with data '%s'", json_data)
                        self._logger.error(e2)
                        error_count += 1

        except Exception as e:
            self._logger.exception(e)

        return error_count, message_count
```

JFSolution:
some problems with rebasing so axel fixed it in this branch:
https://github.com/mrbeam/MrBeamPlugin/pull/1375

octoprint_mrbeam/iobeam/laserhead_handler.py
amrsoll: you can also have a look at https://stackoverflow.com/a/26199212/11136955

JFSolution: > And remove the general try / except as well. You may catch OSError or similar which can happen when the file is not present. You may also catch errors when the Yaml file has incorrect syntax. And to prevent that error from happening, make sure to use `os.path.isfile`.

This was already done

PR #1364, opened 2021-09-23
SW-503 SVG title element breaking the gcode conversion

khaledsherkawi:


amrsoll (approved):
:joy: @hungerpirat we are finally doing the fix!

PR #1369, opened 2021-09-27
Bugfix/sw 517 change label printed to dc s

JFSolution:
moved some text on the printed label


JFSolution:
duplicate with https://github.com/mrbeam/MrBeamPlugin/pull/1370

PR #1373, opened 2021-09-29
SW-528 / Remove broken migration installing pip config

amrsoll:
the devpi server is not online yet, so no need to have this now.

This was preventing the other migrations from happening.


JFSolution (approved):

JFSolution (changes requested):
please bump the version

PR #1374, opened 2021-09-29
SW-528 Hotfix/v0.9.15

amrsoll:
- Fixes SW-528
- version bump to 0.9.15


octoprint_mrbeam/migrate.py
```diff
@@ -207,13 +207,6 @@ def run(self):
                 ):
                     self.disable_gcode_auto_deletion()
 
-                if self.version_previous is None or self._compare_versions(
-                    self.version_previous,
-                    "0.9.0.2",
```
ahmed-mrbeam, line 212: so , now what will happen to versions before this one if they try to update ?

ahmed-mrbeam (approved):

PR #1375, opened 2021-09-30
Feat/sw 467 unfy model id in analytics

JFSolution:


JFSolution:
had problems with rebase, this is was the previous PR https://github.com/mrbeam/MrBeamPlugin/pull/1358

amrsoll (changes requested):
See unresolved conversations in #1358

octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -330,21 +330,16 @@ def _load_laser_heads_file(self):
...
-                        self._last_used_lh_model_id = data["last_used_lh_model_id"]
-
-                    if "correction_settings" in data:
-                        self._correction_settings = data["correction_settings"]
-            except:
-                self._logger.error(
+                    self._lh_cache = data.get("laser_heads")
+                    self._last_used_lh_serial = data.get("last_used_lh_serial")
+                    self._last_used_lh_model_id = data.get("last_used_lh_model_id")
+                    self._correction_settings = data.get("correction_settings")
```
amrsoll, line 336: I like this, we just have to be careful because the values might revert back to `None`, which I find OK but is different logic than before

amrsoll (approved):

octoprint_mrbeam/iobeam/laserhead_handler.py
JFSolution: I checked, the only thing that can be overwritten is the lh_cache all the others are only set here

PR #1385, opened 2021-10-25
SW-568 Undefined var: VERSION_UPDATE_OCTOPRINT_PRERELEASE_FIX

khaledsherkawi:


JFSolution:
I saw, that the description of the migration function is not changed, can you change it so it descirbes what it does
See https://github.com/mrbeam/MrBeamPlugin/commit/2e349f04d0ea387ee3c7c1689baaeca928f90579 where it was merged to beta

JFSolution:
for example """ removes the prerelease flag from the octoprint update config so it will only use releases of octoprint """

JFSolution (approved):

PR #1382, opened 2021-10-21
SW-565 SW-597 added more logging for login issue and added triggerData to analytics

JFSolution:
example anaytics output
```
{
  "sw_tier": "BETA",
  "ntp": false,
  "sid": "5e9819fb9e9b4b3db8012263272e6873@000000008025FBB0-2R",
  "version_mrbeam_plugin": "0.10.0",
  "snr": "000000008025FBB0-2R",
  "data": {
    "browser_time": "20/10/2021, 16:58:49",
    "triggerUrlCount": {
      "/api/plugin/mrbeam": 4
    },
    "error401Count": 4,
    "ts": 1634741929907,
    "triggerData": {
      "/api/plugin/mrbeam": [
        "{\"command\":\"messages\"}",
        "{\"command\":\"messages\"}",
        "{\"command\":\"material_settings\"}",
[...]

PR #1387, opened 2021-10-28
SW-597 add more logging for login issue

JFSolution:
see original PR https://github.com/mrbeam/MrBeamPlugin/pull/1382/files

### DIFF
--- a/octoprint_mrbeam/static/js/mrbeam.js
+++ b/octoprint_mrbeam/static/js/mrbeam.js
@@ -274,7 +274,7 @@ $(function () {
 
         self.passiveLoginInProgress = false;
         self.error401Count = 0;
-        self.triggerUrlCount = {};
+        self.triggerUrl = {};
 
         // This extender forces the input value to lowercase. Used in loginsreen_viewmode.js and wizard_acl.js
         window.ko.extenders.lowercase = function (target, option) {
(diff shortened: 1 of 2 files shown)


octoprint_mrbeam/static/js/mrbeam.js
```diff
@@ -450,27 +452,29 @@ $(function () {
...
-                    self.triggerUrlCount[triggerUrl] = 0;
+                if (!(triggerUrl in self.triggerUrl)) {
+                    self.triggerUrl[triggerUrl] = {'count': 0, 'data': []};
                 }
-                self.triggerUrlCount[triggerUrl]++;
+                self.triggerUrl[triggerUrl]['count']++;
+                self.triggerUrl[triggerUrl]['data'].push(requestData);
+
                 if (self.error401Count === 1) {
                     setTimeout(() => {
```
Yelko-mrbeam, line 465: Just out of curiosity. Why is this on a timeout? was there some bug or is there any other reason? Would be nice to have it documented 😄

Yelko-mrbeam (approved):
Code looks good from my side 🚀

octoprint_mrbeam/static/js/mrbeam.js
JFSolution: The idea is here to send the data in a errorcase only every 2 seconds, so we don't spam analytics

Yelko-mrbeam: But from the code it seems it just waits 2 seconds to send the error after the error occurred or? So in theory you could have two errors one after the other and you will spam analaytics any way only with a 2 second delay 🤔 am I missing something? _handle_session_expired seems to be only called onUserLoggedIn.
In any case not relevant for merging this PR. Just wanted to understand what's happening...

JFSolution: But the timeout is only started on the first apperance of a 401 error, so it will collect all errors that appear in these 2 seconds and then send it and set back the error count to 0 so a new timeout can be started on the next 401

Yelko-mrbeam: ahaa! ok now I get what's going on thanks! Seems a bit obscure from my side as a newby but ok 😄

khaledsherkawi (approved):

PR #1388, opened 2021-11-11
SW-658 merge generated profile at load_all

JFSolution:


octoprint_mrbeam/printing/profile.py
```diff
@@ -257,7 +258,10 @@ def exists(self, identifier):
     def _load_all(self):
         """Extend the file based ``PrinterProfileManager._load_all`` with the few hardcoded ones we have."""
         file_based_profiles = PrinterProfileManager._load_all(self)
-        return dict_merge(LASER_PROFILE_MAP, file_based_profiles)
+        device_type = deviceInfo().get_type()
+        mrbeam_generated_profiles = {device_type: self.get(device_type)}
+        mrbeam_profiles = dict_merge(LASER_PROFILE_MAP, mrbeam_generated_profiles)
+        return dict_merge(mrbeam_profiles, file_based_profiles)
```
amrsoll, line 264: aren't you just creating an empty profile? Where does the profile data get generated?

JFSolution: inside the get methode, I added there that if the identifier is not inside the hardcoded a new one is generated from the default one and just the id and name is changed

amrsoll (approved):

PR #1389, opened 2021-11-15
SW-653 mr beam to use update config from cloud

JFSolution:
this feature enables us to use the update config from https://github.com/mrbeam/beamos_config/blob/master/docs/sw-update-conf.json to be more flexible in changing the config


octoprint_mrbeam/software_update_information.py
```diff
@@ -19,72 +24,197 @@
...
 BEAMOS_LEGACY_DATE = date(2018, 1, 12)
 
+_softwareupdate_handler = None
+
+class MrBeamSoftwareupdateHandler:
+    CLOUD_FILE = 1
+    LOCAL_FILE = 2
+    REPO_FILE = 3
+    def __init__(self, plugin):
+        self._plugin = plugin
```
amrsoll: 1. You do not seem to make the best use of the class structure. Are you sure it is the best method to implement the feature?

2. It does not seem you need the `plugin` object, but rather a subset of information (the selected SW channel). Do you foresee any scenario where you need to use the plugin item itself?

octoprint_mrbeam/software_update_information.py
```diff
@@ -19,72 +24,197 @@
...
+            servererror = True
+            returncode = self.LOCAL_FILE
+
+        # if both invalid use repo file
+        if servererror and localfilemissing:
+            _logger.info(
+                "fallback use local default config file /files/software_update/update_info.json"
+            )
+            try:
+                copy(
```
amrsoll: please write `shutil.copy` as it is less confusing (`copy()` is usually an object copy, from the `copy` package.)

octoprint_mrbeam/software_update_information.py
```diff
@@ -19,72 +24,197 @@
...
+            # inform SoftwareUpdate Pluging about new config
+            sw_update_plugin = self._plugin._plugin_manager.get_plugin_info(
+                "softwareupdate"
+            ).implementation
+            sw_update_plugin._refresh_configured_checks = True
+            sw_update_plugin._version_cache = dict()
+            sw_update_plugin._version_cache_dirty = True
+        return returncode
+
+@logme(output=True)
```
amrsoll: Do you still need it?

octoprint_mrbeam/software_update_information.py
```diff
@@ -19,72 +24,197 @@
...
+class MrBeamSoftwareupdateHandler:
+    CLOUD_FILE = 1
+    LOCAL_FILE = 2
+    REPO_FILE = 3
+    def __init__(self, plugin):
+        self._plugin = plugin
+
+    # @get_thread(daemon=False)
+    @logExceptions
+    def load_update_file_from_cloud(self, localfilemissing=False):
```
amrsoll: What is the benefit of the `localfilemissing` variable ? Because the function checks for the file anyway, so it might as well take care of it itself.

I also think this function is trying to do too many things at the same time: downloading a file and managing local copies.

I would recommend to have different functions that
- tries to download the file and can raise an exception
- verify the presence and content of a local file path
- a function that uses the 2 previous ones to dictate the usage flow : if download fails, use the local copy etc...

This will also help you for the testing so variables like `localfilemissing` are more natural and you have more flexibility.

octoprint_mrbeam/software_update_information.py
```diff
@@ -19,72 +24,197 @@
...
 GLOBAL_PIP_BIN = "/usr/local/bin/pip"
 GLOBAL_PIP_COMMAND = (
     "sudo {}".format(GLOBAL_PIP_BIN) if os.path.isfile(GLOBAL_PIP_BIN) else None
 )
-# GLOBAL_PIP_COMMAND = "sudo {} -m pip".format(GLOBAL_PY_BIN) if os.path.isfile(GLOBAL_PY_BIN) else None #  --disable-pip-version-check
-# VENV_PIP_COMMAND = ("%s -m pip --disable-pip-version-check" % VENV_PY_BIN).split(' ') if os.path.isfile(VENV_PY_BIN) else None
+
 BEAMOS_LEGACY_DATE = date(2018, 1, 12)
 
+_softwareupdate_handler = None
```
amrsoll: If you are going to use a singleton, then please explain (docstring) this variable

octoprint_mrbeam/software_update_information.py
```diff
@@ -19,72 +24,197 @@
...
-            )
-        )
-    return res
+    _logger.info("SoftwareUpdate using tier: %s %s", tier, beamos_date)
+    global _softwareupdate_handler
+    if _softwareupdate_handler is None:
+        _softwareupdate_handler = MrBeamSoftwareupdateHandler(plugin)
+        _softwareupdate_handler.load_update_file_from_cloud()
+
+    return _set_info_from_file(plugin, tier, beamos_date, _softwareupdate_handler)
```
amrsoll: please do not use singletons. You can just create a new instance, there is no benefit to have a singleton here.

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,233 +228,190 @@ def switch_software_channel(plugin, channel):
...
-    # Commented pluginInfo -> If the module is not installed, then it Should be.
-    pluginInfo = plugin._plugin_manager.get_plugin_info(module_id)
-    if pluginInfo is None:
-        display_version = None
+def _set_info_from_file(plugin, tier, beamos_date, _softwareupdate_handler):
+    """
+    loads update info from the update_info.json file
+    the override order: default_settings->module_settings->tier_settings->beamos_settings
+    and if there are update_settings set in the config.yaml they will replace all of the module
+    the json file should look like:
```
amrsoll: I think the name is misleading. It sounds like a function that would take a `file` or a `path` variable, and simply parse the content.

I would recommend to have a parser function separately and some other function that takes care of providing that file, use the parser and return the result to the OctoPrint hook.

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,233 +228,190 @@ def switch_software_channel(plugin, channel):
...
+                sw_update_config[module_id] = moduleconfig
+
+        _logger.debug("sw_update_config {}".format(sw_update_config))
+        return sw_update_config
+
+def clean_update_config(update_config):
+    pop_list = ["alpha", "beta", "stable", "develop", "beamos_date", "name"]
+    for pop_element in pop_list:
+        if pop_element in update_config:
+            update_config.pop(pop_element)
```
amrsoll: you could also use the `del` keyword.
```python
del update_config[pop_element]
```

setup.py
```diff
@@ -48,6 +48,7 @@
     "pillow",
     "lxml",
     "numpy",
+    "yaml",
```
amrsoll: from now on we should try to freeze the dependencies if you could

tests/softwareupdate/test_cloud_config.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def get_plugin_info(self, module_id):
+        return self
+
+    # sw_update_plugin = self._plugin._plugin_manager.get_plugin_info(
+    #     "softwareupdate"
+    # ).implementation
+    # sw_update_plugin._refresh_configured_checks = True
+    # sw_update_plugin._version_cache = dict()
+    # sw_update_plugin._version_cache_dirty = True
```
amrsoll: raus bitte :)

amrsoll (commented):
Why is the update config in the code? We can preload it in the BeamOS filesystem if we want to have it on the devices that ship offline. And we can add it to the usb sticks in the update module. I don't think it is the best idea to hardcode this config.

I cannot see where you parse the "dependencies" key from the update config file, did I miss it? Or we don't use it yet?

octoprint_mrbeam/software_update_information.py
JFSolution: but then it can run multiple processes that will fetch the file from the server, this will make sure that only one process is running that checks the file

octoprint_mrbeam/software_update_information.py
JFSolution: 1. I will check, I think there was something why I choose the class structure
2. plugin will be used to get the path to the local saved file

octoprint_mrbeam/software_update_information.py
JFSolution: ok

octoprint_mrbeam/software_update_information.py
JFSolution: no the logme can be removed before the merge, I only needed it for debuging

octoprint_mrbeam/software_update_information.py
JFSolution: This will be used to force to ignore the local file and use the server file, ok I seperate them

octoprint_mrbeam/software_update_information.py
JFSolution: this is not working

tests/softwareupdate/test_cloud_config.py
JFSolution: done

octoprint_mrbeam/software_update_information.py
JFSolution: I don't see why i should seperate this, it will only open the file catch some exceptions and the parse the content

octoprint_mrbeam/software_update_information.py
amrsoll: Please use a lock if you want to manage threaded access.

octoprint_mrbeam/software_update_information.py
amrsoll: Please use a(n) `(R)Lock` if you want to prevent simultaneous file access. 

Using singletons is more of a coding style for property managers, which I don't think is necessarily good. They carry a baggage of globalised side effects and their influence on the behaviour of other components is difficult to test because they are often coded to be mutable like the settings manager.

octoprint_mrbeam/software_update_information.py
amrsoll: sorry my bad the `del` keyword confuses the python list comprehension. I corrected my comment.

`del` is more specific for removing pointers.

amrsoll: but you can also make it simpler with
```python
for key in set(update_config).intersection(pop_list):
  del update_config[key]
```

octoprint_mrbeam/software_update_information.py
JFSolution: ok

octoprint_mrbeam/software_update_information.py
amrsoll: sure, but then you could just receive the path if it's the only info you need, not the whole plugin :)

octoprint_mrbeam/software_update_information.py
JFSolution: for this methode is no singelton used

octoprint_mrbeam/software_update_information.py
JFSolution: is it done correlty now?

octoprint_mrbeam/software_update_information.py
JFSolution: is also used to get the versionnumber of the octoprint plugins, so I will keep the plugin

octoprint_mrbeam/software_update_information.py
JFSolution: as discussed in the sync meeting, i will use the same lock as for the loading of the file

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,59 +43,188 @@
...
-                _set_info_mrb_hw_info(plugin, tier, beamos_date),
-                # _config_octoprint(plugin, tier),
-            ],
+    _logger.info("SoftwareUpdate using tier: %s %s", tier, beamos_date)
+
+    if plugin._connectivity_checker.check_immediately():
+        config_tag = get_tag_of_github_repo("beamos_config")
+        # if plugin._connectivity_checker.check_immediately():  # check if device online
+        if config_tag:
+            cloud_config = get_config_of_tag(get_tag_of_github_repo("beamos_config"))
```
Yelko-mrbeam: We could directly use config_tag right? Instead of calling get_tag_of_github_repo again

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,59 +43,188 @@
...
+            ]
+            majorversion = Spec(
+                "<" + str(MAJOR_VERSION_CLOUD_CONFIG + 1) + ".0.0"
+            )  # simpleSpec("0.*.*")
+            print(versionlist)
+            return majorversion.select(versionlist)
+        else:
+            _logger.warning("no valid response for the tag of the update_config file")
+            return None
+    except requests.ReadTimeout:
```
Yelko-mrbeam: is this thrown after the 3rd retry if there is another timeout? I would have expected some RetryException 🤔

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,59 +43,188 @@
...
 
+def get_config_of_tag(tag):
+    import requests
+    import json
+
+    try:
+        url = (
+            "https://api.github.com/repos/mrbeam/beamos_config/contents/docs/sw-update-conf.json"
+            + "?ref=v"
+            + str(tag)
```
Yelko-mrbeam: what does this "ref" query param imply?

octoprint_mrbeam/__init__.py
```diff
@@ -966,6 +973,11 @@ def on_wizard_finish(self, handled):
 
     # map(lambda m: m(handled), self._get_subwizard_attrs("_on_", "_wizard_finish").values())
 
+    @octoprint.plugin.BlueprintPlugin.route("/fetch_update_info", methods=["GET"])
```
Yelko-mrbeam: I would encourage creating a new file like update_handler.py under a package called rest_handlers like I did on this [PR](https://github.com/mrbeam/MrBeamPlugin/pull/1415/files). This would allow us to gather all update endpoint related functionality there. Right now it's only one endpoint but it might grow...

Also I would try to follow a more domain driven pattern to make it easier to add functionality like .route('/update/info'). As far as I understand, this endpoint is actually deleting some cache right? if this is the case I would actually vote for making this a DELETE request instead of a GET one.

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,242 +238,209 @@ def switch_software_channel(plugin, channel):
...
-        api_user="MrBeamDev",
-        api_password="v2T5pFkmdgDqbFBJAqrt",
-        pip="git+ssh://git@bitbucket.org/mrbeam/mrb_hw_info.git@{target_version}",
-    )
+    # fmt: off
+    sw_update_plugin = plugin._plugin_manager.get_plugin_info("softwareupdate").implementation
+    # fmt: on
+    sw_update_plugin._refresh_configured_checks = True
+    sw_update_plugin._version_cache = dict()
+    sw_update_plugin._version_cache_dirty = True
```
Yelko-mrbeam: It's not great that we need to rely on the private properties... But I guess there is no otherway so far 😿

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,59 +43,188 @@
...
-    result = dict()
 
+def get_update_information(plugin):
+    """
+    Gets called from the octoprint.plugin.softwareupdate.check_config Hook from Octoprint
+    Starts a thread to look online for a new config file
+    sets the config for the Octoprint Softwareupdate Plugin with the data from the config file
+    @param plugin: calling plugin
+    @return: the config for the Octoprint embedded softwareupdate Plugin
+    """
```
Yelko-mrbeam: Nice doc! Not only here but in all public methods! 👏 ❤️

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,59 +43,188 @@
...
-            )
-        )
-    return res
+
+    sw_update_plugin = plugin._plugin_manager.get_plugin_info(
+        "softwareupdate"
+    ).implementation
+
+    sw_update_plugin._version_cache = dict()
+    sw_update_plugin._version_cache_dirty = True
```
Yelko-mrbeam: what does this part of the code mean? does it make sense to separate it in a method? it seems similar to the one you use at reload_update_info. We are using private properties and we might be able to change this in the future so it might be a good idea to wrap it in a descriptive method. What do you think?

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,242 +238,209 @@ def switch_software_channel(plugin, channel):
...
+        defaultsettings = cloud_config.get("default", None)
+        modules = cloud_config["modules"]
+
+        # TODO maybe drop the higher levels of the config so the output will be more tidy
+        for module_id, module in modules.items():
+            if tier in [
+                SW_UPDATE_TIER_BETA,
+                SW_UPDATE_TIER_DEV,
+                SW_UPDATE_TIER_PROD,
+                SW_UPDATE_TIER_ALPHA,
```
Yelko-mrbeam: Might make sense to make this values an Enum?

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,242 +238,209 @@ def switch_software_channel(plugin, channel):
...
+    @param beamos_date: the image creation date of the running beamos
+    @param _softwareupdate_handler: the handler class to look for a new config file online
+    """
+    if cloud_config:
+        sw_update_config = dict()
+        _logger.debug("update_info {}".format(cloud_config))
+        defaultsettings = cloud_config.get("default", None)
+        modules = cloud_config["modules"]
+
+        # TODO maybe drop the higher levels of the config so the output will be more tidy
```
Yelko-mrbeam: What is this TODO about? shall we create a ticket for it and link it to the comment or should we just do it before merging?

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,242 +238,209 @@ def switch_software_channel(plugin, channel):
...
+
+        if tierversion in input_moduleconfig:
+            input_moduleconfig = dict_merge(
+                input_moduleconfig, input_moduleconfig[tierversion]
+            )  # set tier config from default settings
+
+        if tierversion in input_moduleconfig:
+            input_moduleconfig = dict_merge(
+                input_moduleconfig, input_moduleconfig[tierversion]
+            )  # override tier config from tiers set in config_file
```
Yelko-mrbeam: this lines are exactly the same as the previous ones except for the comment right? is this correct? 😕

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,242 +238,209 @@ def switch_software_channel(plugin, channel):
...
+                    beamos_date >= datetime.strptime(date, "%Y-%m-%d").date()
+                    and prev_beamos_date_entry < beamos_date
+                ):
+                    prev_beamos_date_entry = datetime.strptime(date, "%Y-%m-%d").date()
+                    if tierversion in beamos_config:
+                        beamos_config_module_tier = beamos_config[tierversion]
+                        beamos_config = dict_merge(
+                            beamos_config, beamos_config_module_tier
+                        )  # override tier config from tiers set in config_file
+                    input_moduleconfig = dict_merge(input_moduleconfig, beamos_config)
```
Yelko-mrbeam: can we extract this to a separate self-explanatory method?

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,242 +238,209 @@ def switch_software_channel(plugin, channel):
...
+
+        if "branch" in input_moduleconfig and "{tier}" in input_moduleconfig["branch"]:
```
Yelko-mrbeam: I think it makes sense to have a private method for each if on this public method. I think it would make it more readable and we would follow the single responsibility principle. What do you think @Josef-MrBeam ?

If we decide to go for it I would suggest to separate the methods so that they return what you need and you modify the input_moduleconfig in this public method.

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,242 +238,209 @@ def switch_software_channel(plugin, channel):
...
-        kwargs.update(dict(displayVersion=version))
-
-    return _get_package_description(
-        module_id=module_id, tier=tier, pip_command=pip_command, **kwargs
-    )
+    if tier in [
+        SW_UPDATE_TIER_BETA,
+        SW_UPDATE_TIER_DEV,
+        SW_UPDATE_TIER_PROD,
+        SW_UPDATE_TIER_ALPHA,
```
Yelko-mrbeam: Same Enum could be used here

octoprint_mrbeam/static/js/software_channel_selector.js
```diff
@@ -117,9 +127,18 @@ $(function () {
             button.addClass("sticky-footer");
         };
     }
+    function reload_update_info() {
+        OctoPrint.get("plugin/mrbeam/fetch_update_info")
```
Yelko-mrbeam: if we change to delete and also change the endpoint the change should also be reflected here

octoprint_mrbeam/static/js/software_channel_selector.js
```diff
@@ -117,9 +127,18 @@ $(function () {
...
+            .done(function (response) {
+            })
+            .fail(function () {
+                console.error("Unable to relaod update info.");
```
Yelko-mrbeam: When would it fail? any case we should inform the user?

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,59 +43,188 @@
...
+                )
+    else:
+        _logger.warn("no internet connection")
+
+    user_notification_system = plugin.user_notification_system
+    user_notification_system.show_notifications(
+        user_notification_system.get_notification(
+            notification_id="missing_updateinformation_info", replay=False
         )
     )
```
Yelko-mrbeam: should this be part of the else branch if no internet connection or how does this work? 😕

setup.py
```diff
@@ -44,10 +44,12 @@
...
-    "webcolors",
-    "pillow",
-    "lxml",
-    "numpy",
+    "webcolors == 1.10",
+    "pillow == 6.2.2",
+    "lxml == 4.6.3",
+    "numpy == 1.16.6",
+    "pyyaml == 3.10",
+    "enum == 0.4.7",
```
Yelko-mrbeam: is this working with python 2.7? I had to use enum34==1.1.10 since something wasn't working with this... I don't remember exactly what was it. But since I added it for the MrBeamDoc and both of them provide the same package name I am not sure how will it behave... 
https://github.com/mrbeam/MrBeamDoc/pull/10/files

tests/softwareupdate/test_cloud_config.py
```diff
@@ -0,0 +1,668 @@
...
+                "https://api.github.com/repos/mrbeam/beamos_config/tags",
+                json={"test": "test"},
+                status_code=404,
+            )
+            rm.get(
+                "https://api.github.com/repos/mrbeam/beamos_config/contents/docs/sw-update-conf.json?ref=vNone",
+                status_code=404,
+            )
+            tag_of_github_repo = get_tag_of_github_repo("beamos_config")
+            assert tag_of_github_repo == None
```
Yelko-mrbeam: since you are on a TestCase you could use self.assertIsNone https://docs.python.org/3/library/unittest.html#assert-methods 😄

tests/softwareupdate/test_cloud_config.py
```diff
@@ -0,0 +1,668 @@
...
+        # test for all tiers
+        for tier in tiers:
+            cloud_config = get_config_of_tag(get_tag_of_github_repo("beamos_config"))
+            update_config = _set_info_from_cloud_config(
+                plugin, tier, beamos_date, cloud_config
+            )
+            print("config {}".format(update_config))
+            assert (
+                update_config["octoprint"]
+                == target_octoprint_config[get_tier_by_id(tier)]
```
Yelko-mrbeam: seems like formating is not right or is it? you could use self.assertEquals https://docs.python.org/3/library/unittest.html#assert-methods

octoprint_mrbeam/software_update_information.py
```diff
@@ -98,242 +238,209 @@ def switch_software_channel(plugin, channel):
...
 
 
 @logExceptions
-def _get_octo_plugin_description(module_id, tier, plugin, **kwargs):
-    """Additionally get the version from plugin manager (doesn't it do that by default??)"""
-    # Commented pluginInfo -> If the module is not installed, then it Should be.
-    pluginInfo = plugin._plugin_manager.get_plugin_info(module_id)
-    if pluginInfo is None:
-        display_version = None
+def _set_info_from_cloud_config(plugin, tier, beamos_date, cloud_config):
```
Yelko-mrbeam, line 300: why is this a private method? since we are testing it I assume this should be public, I also don't see why this shouldn't be used from the outside...

tests/softwareupdate/test_cloud_config.py
```diff
@@ -0,0 +1,668 @@
+# coding=utf-8
```
Yelko-mrbeam, line 1: Great job on the testing!! 👏 🎉 

could we include maybe one more test for the user_notification_system to verify that is triggered when we wanted to be triggered?

Yelko-mrbeam (changes requested):

octoprint_mrbeam/software_update_information.py
JFSolution: good point

octoprint_mrbeam/software_update_information.py
JFSolution: with ref you can define the tag of which you want the file from

octoprint_mrbeam/__init__.py
JFSolution: I would not change it to DELETE as it executes the check for updates by clearing the version cache

JFSolution: Ok I will move the blueprint as you described

octoprint_mrbeam/software_update_information.py
Yelko-mrbeam: Ah true.. now that you mention it it seems obvious 😄

octoprint_mrbeam/software_update_information.py
JFSolution: changed the testing to test the public method `get_update_information`

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,27 +1,42 @@
...
-SW_UPDATE_TIER_ALPHA = "ALPHA"
-SW_UPDATE_TIER_DEV = "DEV"
-DEFAULT_REPO_BRANCH_ID = {
-    SW_UPDATE_TIER_PROD: "stable",
-    SW_UPDATE_TIER_BETA: "beta",
-    SW_UPDATE_TIER_ALPHA: "alpha",
-    SW_UPDATE_TIER_DEV: "develop",
-}
+class SWUpdateTier(Enum):
+    PROD = "PROD"
```
Yelko-mrbeam: shouldn't this be STABLE? 😕

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,27 +1,42 @@
...
+    BETA = "BETA"
+    ALPHA = "ALPHA"
+    DEV = "DEV"
+
 
-# add to the display name to modules that should be shown at the top of the list
-SORT_UP_PREFIX = " "
+SW_UPDATE_TIERS_DEV = [SWUpdateTier.ALPHA, SWUpdateTier.DEV]
+SW_UPDATE_TIERS_STABLE = [SWUpdateTier.PROD, SWUpdateTier.BETA]
```
Yelko-mrbeam: and this SW_UPDATE_TIERS_PROD?

Yelko-mrbeam: I really like this idea to have them grouped here and detached from the rest of the code 😄

octoprint_mrbeam/software_update_information.py
```diff
@@ -1,27 +1,42 @@
...
+SW_UPDATE_TIERS = SW_UPDATE_TIERS_DEV + SW_UPDATE_TIERS_STABLE
 
+DEFAULT_REPO_BRANCH_ID = {
```
Yelko-mrbeam, line 32: https://stackoverflow.com/questions/54793033/python-tuple-and-enum/70325042#70325042
this might be interesting for this... we could just add the branch as a property of the enum maybe, what do you think @Josef-MrBeam ?

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,492 @@
...
+        }
+
+        s = requests.Session()
+        retry = Retry(connect=3, backoff_factor=0.3)
+        adapter = HTTPAdapter(max_retries=retry)
+        s.mount("https://", adapter)
+        s.keep_alive = False
+
+        response = s.request("GET", url, headers=headers)
+    except requests.ReadTimeout:
```
Yelko-mrbeam: I think here the same MaxRetryError as before should be catched right?

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,492 @@
...
 )
-# GLOBAL_PIP_COMMAND = "sudo {} -m pip".format(GLOBAL_PY_BIN) if os.path.isfile(GLOBAL_PY_BIN) else None #  --disable-pip-version-check
-# VENV_PIP_COMMAND = ("%s -m pip --disable-pip-version-check" % VENV_PY_BIN).split(' ') if os.path.isfile(VENV_PY_BIN) else None
+
 BEAMOS_LEGACY_DATE = date(2018, 1, 12)
 
 
-def get_update_information(plugin):
-    result = dict()
+def _get_tag_of_github_repo(repo):
```
Yelko-mrbeam: This could be a public method right? 🤔

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,492 @@
...
+        _logger.warning("timeout while trying to get the tag of the update_config file")
+        return None
+    except ConnectionError:
+        _logger.warning(
+            "connection error while trying to get the tag of the update_config file"
+        )
+        return None
+
+
+def _get_config_of_tag(tag):
```
Yelko-mrbeam, line 110: This one also could be public I think... 🤔 or why are they private? the do have a documentation and I think they could be used from another class if needed... What speaks against it? @Josef-MrBeam ? we can leave both as they are, your call 😄 but it would be easier to create smaller unit tests just for this methods

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,492 @@
...
+        tier: software tier
+        beamos_date: date of the beamos
+        plugin: Mr Beam Plugin
+
+    Returns:
+        software update informationen for the module
+    """
+    if tier in SW_UPDATE_TIERS:
+        # merge default settings and input is master
+        input_moduleconfig = dict_merge(defaultsettings, input_moduleconfig)
```
Yelko-mrbeam, line 385: should we rename the assignation to something more accurate like updated_module_configuration? I am guessing the prefix input_ comes from being a parameter? or does it make sense to keep it? @Josef-MrBeam ?
`
        moduleconfig_updated = dict_merge(defaultsettings, input_moduleconfig)
`

tests/softwareupdate/test_cloud_config.py
```diff
@@ -0,0 +1,456 @@
...
+        return self.tier
+
+    def set(self, tier):
+        self.tier = tier
+
+    def settings(self, init=False, basedir=None, configfile=None):
+        return None
+
+
+class DummyConnectifityChecker:
```
Yelko-mrbeam: Connectifity? 😋 is that a typo?

tests/softwareupdate/test_cloud_config.py
```diff
@@ -0,0 +1,456 @@
...
+    _event_bus = EventManager()
+
+    @patch("octoprint.settings.settings")
+    def __init__(self, settings_mock):
+        settings_mock.return_value = None
+        self._logger = mrb_logger("test.Plugindummy")
+        self.user_notification_system = user_notification_system(self)
+
+
+@ddt
```
Yelko-mrbeam: you are not using this in the end right?

Yelko-mrbeam (changes requested):

Yelko-mrbeam:
everything looks good just missing one except MaxRetryError and some potential typos. Great job on the testing!

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,492 @@
...
+    Args:
+        module_id: the id of the software module
+        input_moduleconfig: moduleconfig
+        defaultsettings: default settings
+        tier: software tier
+        beamos_date: date of the beamos
+        plugin: Mr Beam Plugin
+
+    Returns:
+        software update informationen for the module
```
Yelko-mrbeam: informationen -> information/informations

tests/softwareupdate/test_cloud_config.py
JFSolution: no, I will delete it

octoprint_mrbeam/software_update_information.py
JFSolution: yes you are right

octoprint_mrbeam/software_update_information.py
JFSolution: yes you are right

octoprint_mrbeam/software_update_information.py
JFSolution: But if I change it now I need additional migration  :sweat_smile:

JFSolution: I changed the name to STABLE but kept the value

octoprint_mrbeam/software_update_information.py
JFSolution: I checked and need to change to much to get it working and don't see to much benefit

octoprint_mrbeam/software_update_information.py
JFSolution: you are right

octoprint_mrbeam/software_update_information.py
JFSolution: for this one I would keep it as it is as it will get the content of the sw-update-conf file for a given tag

octoprint_mrbeam/software_update_information.py
JFSolution: yes this was used as it is the input through a parameter, so I would keep it

Yelko-mrbeam (approved):

Yelko-mrbeam:
@Josef-MrBeam conflicts need to be resolved

JFSolution:
> @Josef-MrBeam conflicts need to be resolved

I will rebase on the current develop

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,493 @@
...
+
+    return _set_info_from_cloud_config(
+        plugin,
+        tier,
+        beamos_date,
+        {
+            "default": {},
+            "modules": {
+                "mrbeam": {
+                    "name": " MrBeam Plugin offline2",
```
JFSolution: remove offline2

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,493 @@
...
+        json_data = json.loads(response.text)
+        yaml_file = base64.b64decode(json_data["content"])
+
+        return yaml.safe_load(yaml_file)
+    else:
+        _logger.warning("no valid response for the update_config file")
+        return None
+
 
+def get_update_information(plugin):
```
JFSolution, line 155: add try except so the plugin won't break -> log to analytics

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,493 @@
...
+                sw_update_config[module_id] = _generate_config_of_module(
+                    module_id, module, defaultsettings, tier, beamos_date, plugin
+                )
+
+        _logger.debug("sw_update_config {}".format(sw_update_config))
+
+        sw_update_file_path = os.path.join(
+            plugin._settings.getBaseFolder("base"), SW_UPDATE_INFO_FILE_NAME
+        )
+        try:
```
JFSolution, line 349: check if we need to except more errors, for example for dumps

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,493 @@
...
+            with open(sw_update_file_path, "w") as f:
+                f.write(json.dumps(sw_update_config))
+        except IOError:
+            plugin._logger.error("can't create update info file")
```
JFSolution, line 353: add this to analytics

octoprint_mrbeam/software_update_information.py
```diff
@@ -32,308 +47,493 @@
...
+        repo: repository name
+
+    Returns:
+        latest tag of the given majorversion <MAJOR_VERSION_CLOUD_CONFIG>
+    """
+    import requests
+    import json
+
+    try:
+        url = "https://api.github.com/repos/mrbeam/" + repo + "/tags"
```
JFSolution: change to `"https://api.github.com/repos/mrbeam/{repo}/tags".format(repo=repo)`

octoprint_mrbeam/static/js/software_channel_selector.js
```diff
@@ -117,9 +127,18 @@ $(function () {
             button.addClass("sticky-footer");
         };
     }
+    function reload_update_info() {
+        OctoPrint.post("plugin/mrbeam/info/update")
+            .done(function (response) {
+            })
+            .fail(function () {
+                console.error("Unable to reload update info.");
```
JFSolution: add analytics

setup.py
```diff
@@ -44,10 +44,12 @@
 else:
     picamera = 'picamera; platform_machine=="armv7l"'
 plugin_requires = [
-    "webcolors",
-    "pillow",
-    "lxml",
-    "numpy",
+    "webcolors <= 1.10",
```
JFSolution: move this definition of the versions to a seperate issue

setup.py
```diff
@@ -82,6 +84,7 @@
             "files/migrate_logrotate/*",
             "files/material_settings/*",
             "files/camera/*",
+            "files/software_update/*",
```
JFSolution: this is not needed anymore, please remove

octoprint_mrbeam/migrate.py
```diff
@@ -49,6 +52,7 @@ class Migration(object):
     VERSION_UPDATE_CUSTOM_MATERIAL_SETTINGS = "0.9.9"
     VERSION_UPDATE_OCTOPRINT_PRERELEASE_FIX = "0.9.10"
     VERSION_UPDATE_FORCE_FOCUS_REMINDER = "0.10.0"
+    VERSION_ENABLE_ONLINECHECK = "0.11.0"
```
JFSolution: remove

octoprint_mrbeam/migrate.py
```diff
@@ -404,6 +412,15 @@ def save_current_version(self):
             )  # force needed to save it if it wasn't there
             self.plugin._settings.save()
 
+    @property
+    def restart(self):
+        return self._restart
+
+    @restart.setter
+    def restart(self, value):
+        if self._restart == 0 or value < self._restart:
```
JFSolution, line 421: change so None won't override the restart

octoprint_mrbeam/__version.py
```diff
@@ -1 +1 @@
-__version__ = "0.10.3"
+__version__ = "0.11.0"
```
JFSolution: revert

octoprint_mrbeam/software_update_information.py
JFSolution: @ahmed-mrbeam errors will already be added to analytics

PR #1398, opened 2021-11-23
Do not handle unicode exception if not needed.

amrsoll:
In the log function, the TypeError that is removed here will
take precedent over an other potential Exception being logged. - Thereby
we only manage to log the error message, not the stacktrace.

The proposed solution handles the unicode conversion for instances that
we know can be converted.

### DIFF
--- a/octoprint_mrbeam/mrb_logger.py
+++ b/octoprint_mrbeam/mrb_logger.py
@@ -84,11 +84,9 @@ class MrbLogger(object):
         :type kwargs:
         """
 
-        try:
+        import io
+        if isinstance(msg, (str, io.IOBase)):
             msg = unicode(msg, "utf-8")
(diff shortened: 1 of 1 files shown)


octoprint_mrbeam/mrb_logger.py
```diff
@@ -84,11 +84,9 @@ def log(self, level, msg, *args, **kwargs):
         :type kwargs:
         """
 
-        try:
+        import io
+        if isinstance(msg, (str, io.IOBase)):
```
Yelko-mrbeam, line 88: As discussed offline:
- We could add some quick unit tests to make sure this is not crashing if msg is of type Exception for example and whatever other type you think it might be interesting.

Yelko-mrbeam (changes requested):

PR #1402, opened 2021-11-26
SW-742-mrbeam-plugin-init-failed-after-update-to-alpha-0-10-0

ahmed-mrbeam:
Please refer to the issue for more details
https://mr-beam.atlassian.net/browse/SW-742

### DIFF
--- a/octoprint_mrbeam/iobeam/laserhead_handler.py
+++ b/octoprint_mrbeam/iobeam/laserhead_handler.py
@@ -332,7 +332,7 @@ class LaserheadHandler(object):
                 if data:
                     self._lh_cache = data.get("laser_heads")
                     self._last_used_lh_serial = data.get("last_used_lh_serial")
-                    self._last_used_lh_model_id = data.get("last_used_lh_model_id")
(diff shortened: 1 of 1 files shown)


amrsoll (commented):
I think we are mixing up stuff by renaming this the wrong thing (as lh_model and lh_model_id are different)

ahmed-mrbeam:
> I think we are mixing up stuff by renaming this the wrong thing (as lh_model and lh_model_id are different)

@amrsoll explain more

amrsoll:
I understand better. The screenshots fo the text were confusing. What you are looking to use is the lh_model_string perhaps, and you would have to do the reverse mapping to get the ID back

Yelko-mrbeam (approved):

amrsoll:
I think your fix will break because the values are different

amrsoll:
or perhaps there is a migration step, in which case going back to beta from alpha might break

amrsoll (approved):

amrsoll:
For context,. the matter was discussed offline: The key was preserved from previous version to prevent using one more migration step ; and the values get sanitised at an other location in the code.

PR #1403, opened 2021-11-30
SW-742 mrbeam plugin init failed after update to alpha 0 10 0

ahmed-mrbeam:


octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -323,36 +363,49 @@ def _calculate_power_correction_factor(self):
...
+                self._last_used_lh_model = data.get("last_used_lh_model")
+                self._last_used_lh_model_id = data.get("last_used_lh_model_id")
+                self._correction_settings = data.get("correction_settings")
+
+                self._logger.info("Loading data from  {} is successful!".format(self._laser_heads_file))
+            else:
+                self._logger.warn(
+                    "Couldn't load laser head data, file: {} data: {}".format(self._laser_heads_file, data))
+        except IOError as e:
+            self._logger.error("Can't open file: {} {}".format(self._laser_heads_file, e))
```
Yelko-mrbeam, line 385: Will we end up here if the file does not exists?

octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -50,14 +52,29 @@ def __init__(self, plugin):
...
+            self._current_used_lh_model_id, self._current_used_lh_model))
+        laser_head_model_str_list = [v for (k, v) in self._LASERHEAD_MODEL_STRING_MAP.items()
+                                     if ((v == str(self._current_used_lh_model)) or
+                                         (k == str(self._current_used_lh_model_id)))]
+        try:
+            return laser_head_model_str_list.pop(0)
+        except IndexError:
+            self._logger.error("Unknown laserhead model,  name: {} ID: {}".format(
+                self._current_used_lh_model, self._current_used_lh_model_id))
+            return str(None)
```
amrsoll, line 71: Why don't you want to raise the Exception? I thought "None"s were bad to have in our values

octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -323,36 +363,49 @@ def _calculate_power_correction_factor(self):
...
-            last_used_lh_model=self._current_used_lh_model_id,
+            last_used_lh_model=self._current_used_lh_model,
+            last_used_lh_model_id=self._current_used_lh_model_id,
         )
-        file = self._laser_heads_file if file is None else file
+        laser_heads_file = self._laser_heads_file if laser_heads_file is None else laser_heads_file
+        self._logger.info("Writing to file: {} ..started".format(laser_heads_file))
         try:
-            with open(file, "w") as outfile:
+            with open(laser_heads_file, "w+") as outfile:
```
amrsoll: Why would you ~append to read from the file here?~

octoprint_mrbeam/iobeam/laserhead_handler.py
ahmed-mrbeam: in this case we don't need to crash the system because the laserhead string is none, because after iobeam is up. it will have a valid string value, so we log the error and we handle it

octoprint_mrbeam/iobeam/laserhead_handler.py
amrsoll: Sry, why would you want to read the file here?

octoprint_mrbeam/iobeam/laserhead_handler.py
ahmed-mrbeam: supposedly

octoprint_mrbeam/iobeam/laserhead_handler.py
ahmed-mrbeam: changed

octoprint_mrbeam/iobeam/laserhead_handler.py
ahmed-mrbeam: Screenshot from 2021-11-30 16-09-08

amrsoll (approved):

PR #1408, opened 2021-12-27
SW-635 Design Store DEV settings page returning an error while closed login loop happens

hungerpirat:
fixes SW-635

replaced 6 occurences of self.loginState.currentUser?.().settings with a helper method self.getUserSettings() - which uses self.loginState.currentUser?.()?.settings <- with ? behind ().

### DIFF
--- a/octoprint_mrbeam/static/js/design_store.js
+++ b/octoprint_mrbeam/static/js/design_store.js
@@ -21,27 +21,29 @@ $(function () {
             self.prepareDesignStoreTab();
         };
 
+        self.getUserSettings = function () {
+            const settings = self.loginState.currentUser?.()?.settings;
+            return settings;
+        };
+
         self.getEmail = function () {
-            if (
-                self.loginState.currentUser?.().settings?.mrbeam
-                    ?.design_store_email
-            ) {
-                return self.loginState.currentUser().settings.mrbeam
-                    .design_store_email;
+            const userSettings = self.getUserSettings();
+            if (userSettings?.mrbeam?.design_store_email) {
+                return userSettings.mrbeam.design_store_email;
             } else {
                 return self.loginState.username();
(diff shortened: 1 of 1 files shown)


octoprint_mrbeam/static/js/design_store.js
```diff
@@ -209,7 +223,7 @@ $(function () {
                 error: function (jqXHR, textStatus, errorThrown) {
                     console.error(
                         "Store bought design saving failed with status " +
-                        jqXHR.status,
+                            jqXHR.status,
```
Yelko-mrbeam, line 226: There are some formatting differences. Do we have a standard formatting ruleset to verify which one is the correct one?

octoprint_mrbeam/static/js/design_store.js
```diff
@@ -21,44 +21,43 @@ $(function () {
             self.prepareDesignStoreTab();
         };
 
+        self.getUserSettings = function () {
+            const settings = self.loginState.currentUser?.()?.settings;
+            return settings;
+        };
```
Yelko-mrbeam, line 27: Nice refactoring, much more readable

Yelko-mrbeam (approved):

octoprint_mrbeam/static/js/design_store.js
khaledsherkawi: Yea I believe we agreed on "prettier" at one point but it's not used by everyone. I believe we can approve this for now as we will be deciding on new formatters very soon among the team.

khaledsherkawi (approved):

octoprint_mrbeam/static/js/design_store.js
khaledsherkawi: Indeed

PR #1410, opened 2021-12-28
SW-724 drop-upload-either-in-working-area-or-in-design-lib-directly

hungerpirat:


khaledsherkawi (approved):

khaledsherkawi:
@hungerpirat Shall we also restrict the "upload" icon from showing on tabs other than the working area and the design library?
Right now if you drag any file into any other tabs (settings, design store, .. ) the upload icon will still show.

hungerpirat:
@khaledsherkawi no restriction. Basically you're right, this is something to discuss. Here's why I still don't want a restriction: 
1. Iterate in smaller steps. We have already an improvement, bring that on the road. The behaviour on the other tabs stays the same, so it doesn't make things worser. 
2. I could not give a clear answer what the best behaviour on other tabs should be at the time being. This needs more thoughts and should not delay the other improvement.

PR #1415, opened 2021-12-30
SW-827 add docs endpoint to serve the documents comming from MrBeamDoc

Yelko-mrbeam:
https://mr-beam.atlassian.net/browse/SW-827


requirements.txt
```diff
@@ -1,9 +1,77 @@
...
+idna==2.10
+importlib-metadata==2.1.2
+importlib-resources==3.3.1
+itsdangerous==1.1.0
+Jinja2==2.8.1
+Markdown==2.6.11
+MarkupSafe==1.1.1
+monotonic==1.3
+more-itertools==5.0.0
+-e git+git@github.com:mrbeam/MrBeamDoc.git@fe6aeb064dee2d5d987465f5dc251c349a4059af#egg=MrBeamDoc
```
Yelko-mrbeam: Not sure about how to include this on the requirements. We should discuss about this. Also I only added the requirements need to run the test locally so far not the requirements needed to run the current version of our system on a rasberrypi. There are some differences I could find by doing pip freeze on a mr beam.

requirements.txt
```diff
@@ -1,9 +1,77 @@
...
+netaddr==0.7.20
+netifaces==0.10.9
+numpy==1.16.6
+-e git+git@github.com:mrbeam/OctoPrint.git@4336a57578346f2706b2bcb2b5556403d0f7ce85#egg=OctoPrint
```
Yelko-mrbeam: same here

requirements.txt
JFSolution: The requirements of a octoprint plugin have to be in the setup py plugin_requires section

Yelko-mrbeam: Yes. I added it here so that it's easier to run the tests locally. Not really for the project... But it's true that I should add the dependency to the setup.py for the mrbeamdoc at least.

JFSolution: But can this cause problems during install? If it is in multiple locations

Yelko-mrbeam: That was my question for you guys 😓
I think it makes sense to have this somewhere my question is. Does it make trouble to keep it here? should I move it to a different file and leave the requirements file for now? I think we would need to have it in both places or ideally have it here and load it from the setup.py

octoprint_mrbeam/rest_handler/docs_handler.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @octoprint.plugin.BlueprintPlugin.route(
+        "/docs/<string:model>/<string:language>/<string:doctype>.<string:extension>", methods=["GET"])
+    def get_doc(self, model, doctype, language, extension):
+        self._logger.debug(
+            'Request to Model: %(model)s Doctype: %(doctype)s Language: %(language)s Extension:%(extension)s',
+            {'model': model, 'doctype': doctype, 'language': language, 'extension': extension})
+
+        mrbeam_model_found = next(
+            (mrbea_model for mrbea_model in MrBeamModel if mrbea_model.value.lower() == model.lower()), None)
```
JFSolution: is this a typo or intended to be `mrbea_model`?

octoprint_mrbeam/rest_handler/docs_handler.py
```diff
@@ -0,0 +1,36 @@
...
+        supported_language_found = next((supported_language for supported_language in SupportedLanguage if
+                                         supported_language.value.lower() == language.lower()), None)
+        mrbeam_doctype_found = next(
+            (mrbea_doctype for mrbea_doctype in MrBeamDocType if mrbea_doctype.value.lower() == doctype.lower()), None)
```
JFSolution: same here

setup.py
```diff
@@ -49,6 +49,7 @@
     "lxml",
     "numpy",
     picamera,
+    'octoprint_mrbeamdoc',
```
JFSolution: I think this is not working, it will not be found on pypi

octoprint_mrbeam/model/settings_model.py
```diff
@@ -0,0 +1,33 @@
+class SettingsModel:
+    def __init__(self):
+        self.about = None
+
+    def __repr__(self):
+        return 'SettingsModel(about=%s)' % (repr(self.about))
+
+
+class About:
```
Yelko-mrbeam: append model to al these class names

JFSolution (dismissed):

Yelko-mrbeam:
@Josef-MrBeam translations for the documents are missing... I would need help with adding them 😇

JFSolution (dismissed):

JFSolution:
> @Josef-MrBeam translations for the documents are missing... I would need help with adding them 😇

The translation usually will be added during beta release

Yelko-mrbeam:
> > @Josef-MrBeam translations for the documents are missing... I would need help with adding them 😇
> 
> The translation usually will be added during beta release

They are not missing anymore, so all good 😄

JFSolution (approved):

PR #1416, opened 2022-01-11
bugfix/SW-890 svg embedded images are misaligned

hungerpirat:
Way simpler solution than before: just apply total transform matrix instead of modifying attributes x, y, width, height. 
Attribute modification is not able to reflect rotation and mirroring at all. Therefore there is no way to remove the transform completely. Finally it does not make sense to reduce it to mirroring or rotation only.


khaledsherkawi:
This was merged into `feature/SW-990-create-and-test-a-feature-branch-including-the-attached-tickets`

khaledsherkawi:
done via https://github.com/mrbeam/MrBeamPlugin/pull/1480

PR #1423, opened 2022-01-24
SW-724 drop upload either in working area or in design lib directly 3

khaledsherkawi:


octoprint_mrbeam/static/js/mother_viewmodel.js
```diff
@@ -376,7 +376,17 @@ $(function () {
                 }
 
                 if (file) {
-                    self.workingArea.placeUpload(file);
+                    const fileAlreadyPlaced = (element) => element.path === path;
```
JFSolution, line 379: I thought the idea was to replace the file if it was already there

JFSolution (approved):

PR #1428, opened 2022-02-03
Hotfix/SW-974 disable netconnectd debug log

JFSolution:


JFSolution:
open question:
- should we change the rsyslog as it is done for the legacy image?
- what happens for devices in beta? they won't get this migration step

octoprint_mrbeam/__version.py
```diff
@@ -1 +1 @@
-__version__ = "0.10.1"
+__version__ = "0.10.1-hotfix"
```
ahmed-mrbeam: "0.10.1-hotfix.1" ?

octoprint_mrbeam/files/migrate_logrotate/netconnectd
```diff
@@ -6,4 +6,6 @@
     missingok
     notifempty
     copytruncate
+    dateext
+    dateformat %Y-%m-%d
```
ahmed-mrbeam: "." is missing --> .%Y-%m-%d

octoprint_mrbeam/migrate.py
```diff
@@ -71,6 +75,13 @@ def __init__(self, plugin):
         )
         beamos_tier, self.beamos_date = self.plugin._device_info.get_beamos_version()
 
+        # TODO get beamos version of legacy and buster image
+        if os.path.exists("/etc/beamos_version"):
+            with open("/etc/beamos_version") as f:
+                self.beamos_version = f.read()
+        else:
+            self.beamos_version = None
+
```
ahmed-mrbeam: Please move that to a function and this function should be moved to the utils pakcage "octoprint_mrbeam/util/os_utils.py"

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
 
+            self._run_migration()
             self.save_current_version()
         except MigrationException as e:
             self._logger.exception("Error while migration: {}".format(e))
         except Exception as e:
             self._logger.exception("Unhandled exception during migration: {}".format(e))
 
+    def _run_migration(self):
+        """
```
ahmed-mrbeam, line 281: Please follow Google doc string format in all your doc strings

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
+
+        list_of_migrations_available_to_run = [
+            MigrationBaseClass.return_obj(migration, self.plugin)
+            for migration in list_of_migrations
+            if migration.shouldrun(migration, self.beamos_version) is not None
+        ]
+        self._logger.debug(list_of_migrations_available_to_run)
+
+        if not len(list_of_migrations_available_to_run):
+            self._logger.debug("no migration needed")
```
ahmed-mrbeam: self._logger.debug("MIG2.0 > No migration needed ")

The tag at the begging will help us to easily identify the source of this log, you can also include in the logs the method name, file name

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
+            return
+
+        migration_executed_file = os.path.join(
```
ahmed-mrbeam: migrations_json_file_path

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
+            self.plugin._settings.getBaseFolder("base"), "migrations.json"
+        )
+
+        # if file is not there create it
+        if not os.path.exists(migration_executed_file):
+            with open(migration_executed_file, "w") as f:
```
ahmed-mrbeam: Please add an info log about the file creation

ahmed-mrbeam: Are we sure that "w" only is enough to create the file ?

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
                 self._logger.debug("No migration required.")
 
+            self._run_migration()
             self.save_current_version()
         except MigrationException as e:
             self._logger.exception("Error while migration: {}".format(e))
         except Exception as e:
             self._logger.exception("Unhandled exception during migration: {}".format(e))
 
+    def _run_migration(self):
```
ahmed-mrbeam, line 280: how about we move this function to be a static method in the migration base class ?

octoprint_mrbeam/migration/Mig001.py
```diff
@@ -0,0 +1,36 @@
+from octoprint_mrbeam.migration.migration_base import (
+    MigrationBaseClass,
+)
+
+
+class Mig001NetconnectdDisableLogDebugLevel(MigrationBaseClass):
+
```
ahmed-mrbeam, line 10: missing class doc string

octoprint_mrbeam/migrate.py
```diff
@@ -1,21 +1,25 @@
...
 from collections import Iterable, Sized, Mapping
 import os
-import platform
 import re
 import shutil
 from datetime import datetime
 from distutils.version import LooseVersion, StrictVersion
 
 from octoprint_mrbeam import IS_X86
+from octoprint_mrbeam.migration.migration_base import (
```
ahmed-mrbeam: would suggest to move all these imports to the here 
`octoprint_mrbeam/migration/__init__.py`
 and Just use this here instead 
 from octoprint_mrbeam import migration

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
+        if not os.path.exists(migration_executed_file):
+            with open(migration_executed_file, "w") as f:
+                json.dump({}, f)
+
+        try:
+            with open(migration_executed_file, "r") as f:
+                try:
+                    migration_executed = json.load(f)
+                except ValueError:
+                    migration_executed = {}
```
ahmed-mrbeam: I wouldn't do that , this is clearly bypassing the  `ValueError` , what happens if the file for some other reason was corrupted ? with this handling we will assume an empty migration list and continue and then this would fail later.

I would suggest that if this happens, we need to log the error and fail the migration.

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
+                    if migration.id in migration_executed:
+                        if migration_executed[migration.id]:
+                            list_of_migrations_to_run.remove(migration)
+                        else:
+                            # migration failed, should stay in execution queue and the following too
+                            break
+
+                # run migrations
+                for migration in list_of_migrations_to_run:
+                    # run migration
```
ahmed-mrbeam: no need for this comment, the next line is self explanatory

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
             self._logger.exception("Unhandled exception during migration: {}".format(e))
 
+    def _run_migration(self):
+        """
+        run the new migrations
+        @return:
+        """
+        self._logger.debug("beamos_version: " + self.beamos_version)
+
+        list_of_migrations_available_to_run = [
```
ahmed-mrbeam: I would rename that to list_of_migrations_obj_available_to_run

octoprint_mrbeam/migrate.py
```diff
@@ -264,12 +276,74 @@ def run(self):
...
+                json.dump({}, f)
+
+        try:
+            with open(migration_executed_file, "r") as f:
+                try:
+                    migration_executed = json.load(f)
+                except ValueError:
+                    migration_executed = {}
+
+                list_of_migrations_to_run = list(list_of_migrations_available_to_run)
```
ahmed-mrbeam: would rename that to list_of_migrations_obj_to_run

octoprint_mrbeam/migrate.py
```diff
@@ -71,6 +75,13 @@ def __init__(self, plugin):
         )
         beamos_tier, self.beamos_date = self.plugin._device_info.get_beamos_version()
 
+        # TODO get beamos version of legacy and buster image
```
ahmed-mrbeam: remove this pleaes

octoprint_mrbeam/migrate.py
```diff
@@ -71,6 +75,13 @@ def __init__(self, plugin):
...
+        if os.path.exists("/etc/beamos_version"):
```
ahmed-mrbeam: Please add handling for legacy image version as well.. so this function should handle both

octoprint_mrbeam/migrate.py
```diff
@@ -1,21 +1,25 @@
...
 from octoprint_mrbeam.software_update_information import BEAMOS_LEGACY_DATE
 from octoprint_mrbeam.mrb_logger import mrb_logger
 from octoprint_mrbeam.util.cmd_exec import exec_cmd, exec_cmd_output
-from octoprint_mrbeam.util import logExceptions, dict_get
+from octoprint_mrbeam.util import logExceptions
 from octoprint_mrbeam.printing.profile import laserCutterProfileManager
 from octoprint_mrbeam.printing.comm_acc2 import MachineCom
 from octoprint_mrbeam.__version import __version__
 from octoprint_mrbeam.materials import materials
+from migration import list_of_migrations
```
ahmed-mrbeam: Same as above comment

octoprint_mrbeam/migration/migration_base.py
```diff
@@ -0,0 +1,142 @@
...
+from octoprint_mrbeam import mrb_logger
+from octoprint_mrbeam.util.cmd_exec import exec_cmd
+
+
+# TODO build guide in confluence
+
+
+class MIGRATION_STATE(enumerate):
+    init = 1
+    migrate = 2
```
ahmed-mrbeam: `migration_started` ?

octoprint_mrbeam/migration/migration_base.py
```diff
@@ -0,0 +1,142 @@
...
+    migrationDone = 3
+    rollback = 4
```
ahmed-mrbeam: `rollback_started`  ?

octoprint_mrbeam/migration/migration_base.py
```diff
@@ -0,0 +1,142 @@
...
+    rollbackDone = 5
+    error = -1
+
+
+@six.add_metaclass(abc.ABCMeta)
+class MigrationBaseClass:
+    # folder of the files needed during migration
+    MIGRATE_FILES_FOLDER = "files/migrate/"
```
ahmed-mrbeam: Please use `os.path.join()`

ahmed-mrbeam (changes requested):

octoprint_mrbeam/software_update_information.py
```diff
@@ -309,7 +309,9 @@ def _get_package_description(
     if "{tier}" in branch_default:
         branch_default = branch_default.format(tier=get_tier_by_id(tier))
     if prerelease_channel and "{tier}" in prerelease_channel:
-        kwargs.update(prerelease_channel=prerelease_channel.format(tier=get_tier_by_id(tier)))
+        kwargs.update(
```
ahmed-mrbeam: Only format ?, is this format needed ?

octoprint_mrbeam/migration/migration_base.py
```diff
@@ -0,0 +1,142 @@
...
+        this will wrap the migration execution
+        @return:
+        """
+        self._setState(MIGRATION_STATE.migrate)
+        self._logger.info("start migration of " + self.__class__.__name__)
+        try:
+            self._run()
+            self._setState(MIGRATION_STATE.migrationDone)
+            self._logger.info("end migration of " + self.__class__.__name__)
+            if self._state != MIGRATION_STATE.migrationDone:
```
ahmed-mrbeam: Not quit sure how this should work, we set the state and then check it ?

octoprint_mrbeam/migration/migration_base.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def rollback(self):
+        """
+        this will wrap the rollback execution
+        @return:
+        """
+        self._setState(MIGRATION_STATE.rollback)
+        self._logger.warn("start rollback " + self.__class__.__name__)
+        self._rollback()
+        self._setState(MIGRATION_STATE.rollbackDone)
```
ahmed-mrbeam: based on what we decide it is done ? you have to use an evidence , like return code or boolean

octoprint_mrbeam/files/migrate_logrotate/netconnectd
JFSolution: the logrotate will not be done during this migration, I will rollback this change

octoprint_mrbeam/__version.py
JFSolution: ok, will change

ahmed-mrbeam:
> open question:
> 
> * should we change the rsyslog as it is done for the legacy image?
> * what happens for devices in beta? they won't get this migration step

octoprint_mrbeam/migrate.py
JFSolution: we already have a util/device_info, I would put it there

octoprint_mrbeam/migrate.py
JFSolution: This is not releated to the migrations it will only tell if no migration is needed
the source of the log will be `octoprint.plugins.mrbeam.migrate`

ahmed-mrbeam: The thing is , We need to distinguish for which migration generation this belongs to , old way or new way

octoprint_mrbeam/migrate.py
JFSolution: ok

octoprint_mrbeam/migrate.py
JFSolution: I tested it it is working

octoprint_mrbeam/migrate.py
JFSolution: What qould be the benefit? For me it makes more sense here as the migrate.py is reponsible for executing the migrations

octoprint_mrbeam/migration/Mig001.py
JFSolution: will add

octoprint_mrbeam/migrate.py
JFSolution: ok

octoprint_mrbeam/migrate.py
ahmed-mrbeam: The gained benefir is that we have the logic for the new migration process all under one folder, and when you try to run it, you will always use the base class so it would be more clear in the code

octoprint_mrbeam/migrate.py
JFSolution: This doesn't wokr, I would have to import it as 
`from octoprint_mrbeam import migration as mbream_migration`
and then use it as
`mrbeam_migration.MigrationBaseClass` and so on

octoprint_mrbeam/migrate.py
JFSolution: As we now create the file if it is not there we can change it, you are right

octoprint_mrbeam/migrate.py
JFSolution: ok, will remove

octoprint_mrbeam/migrate.py
JFSolution: ok

octoprint_mrbeam/software_update_information.py
JFSolution: no this was some black reformatting and ended up in a commit

octoprint_mrbeam/migration/migration_base.py
JFSolution: indeed this doesn't make sense

octoprint_mrbeam/migration/migration_base.py
JFSolution: if there were no errors during execution it should be successfull

ahmed-mrbeam (dismissed):

ahmed-mrbeam (commented):
waiting for final testing

ahmed-mrbeam (approved):

PR #1430, opened 2022-02-08
SW-974-fix-legacy image

JFSolution:
fix that the migration run for the legacy image
and fix beamos_version refactor bug

### DIFF
--- a/octoprint_mrbeam/migration/Mig001.py
+++ b/octoprint_mrbeam/migration/Mig001.py
@@ -14,6 +14,7 @@ class Mig001NetconnectdDisableLogDebugLevel(MigrationBaseClass):
     def __init__(self, plugin):
         """
         initalization of the migration 001
+
         Args:
             plugin: Mr Beam Plugin
         """
@@ -23,6 +24,7 @@ class Mig001NetconnectdDisableLogDebugLevel(MigrationBaseClass):
     def id(self):
         """
         return the id of the migration
+
         Returns:
             string: id of the migration
         """
@@ -31,6 +33,7 @@ class Mig001NetconnectdDisableLogDebugLevel(MigrationBaseClass):
     def _run(self):
         """
         migration steps executet during migration
+
         Returns:
             None
         """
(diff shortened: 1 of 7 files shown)


ahmed-mrbeam:
I also discovered one more thing. Please see my comment here 
https://github.com/mrbeam/MrBeamPlugin/pull/1429/files

JFSolution:
> I also discovered one more thing. Please see my comment here https://github.com/mrbeam/MrBeamPlugin/pull/1429/files

I can't see your comment

tests/migrations/test-migration-Mig001.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class TestMigrationMig001(unittest.TestCase):
+    """
+    Testclass for the migration Mig001
+    """
+    def setUp(self):
+        self.m001 = Mig001NetconnectdDisableLogDebugLevel(None)
+
+    def test_beamos_versions(self):
```
ahmed-mrbeam, line 12: for the sake of completion, we should add an assert for non matching version string

octoprint_mrbeam/migration/Mig001.py
```diff
@@ -14,6 +14,7 @@ class Mig001NetconnectdDisableLogDebugLevel(MigrationBaseClass):
     def __init__(self, plugin):
         """
         initalization of the migration 001
+
```
ahmed-mrbeam, line 17: would be nice if the  white space formatting here could be reverted

JFSolution: this is needed to show the doc strings correctly, thats why I added it

octoprint_mrbeam/migration/migration_base.py
```diff
@@ -73,13 +75,19 @@ def id(self):
...
         Checks if this Miration should run
+
         Args:
             cls: Migrationclass
             beamos_version: current beamos_version
 
         Returns:
             bool: True if this migration should run
         """
+        if not isinstance(beamos_version, basestring):
```
ahmed-mrbeam, line 86: Very nice selection of the class to compare to .. a link to confirm your good choice  
https://stackoverflow.com/a/1979107

ahmed-mrbeam (approved):

PR #1432, opened 2022-02-10
SW-1046 Modify the "userData" sent to design store by sanitizing the mr beam plugin verison

khaledsherkawi:
SW-1046 Sanitize the Mr Beam version in user data, before it is sent to the Design Store

### DIFF
--- a/octoprint_mrbeam/static/js/design_store.js
+++ b/octoprint_mrbeam/static/js/design_store.js
@@ -124,7 +124,10 @@ $(function () {
                 email: self.getEmail(),
                 serial: MRBEAM_SERIAL,
                 user_token: self.getAuthToken(),
-                version: BEAMOS_VERSION,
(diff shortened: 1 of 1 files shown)


octoprint_mrbeam/static/js/design_store.js
```diff
@@ -124,7 +124,10 @@ $(function () {
                 email: self.getEmail(),
                 serial: MRBEAM_SERIAL,
                 user_token: self.getAuthToken(),
-                version: BEAMOS_VERSION,
+                // TODO: remove the following sanitization once the version
```
ahmed-mrbeam: Please add the issue ID here in the comment for the issue that introduced the change and the issue ID for the one that should revert it

khaledsherkawi: Done

ahmed-mrbeam (approved):

PR #1433, opened 2022-02-14
SW-1048 modify the version comparative methods in the mr beam plugin

khaledsherkawi:
- Add semantic versioning comparison library https://www.npmjs.com/package/compare-versions
- Remove old version comparison methods
- Replace old version comparison methods with the new one
- Remove deprecated method


octoprint_mrbeam/static/js/loginscreen_viewmodel.js
```diff
@@ -18,14 +18,11 @@ $(function () {
         };
 
         self.onStartupComplete = function () {
-            // if (window.mrbeam.isOctoPrintVersionMin("1.3.6")) {
```
Yelko-mrbeam, line 21: nice 👏

octoprint_mrbeam/static/js/lib/compare-versions.js
```diff
@@ -0,0 +1,165 @@
+/* global define */
+(function (root, factory) {
```
Yelko-mrbeam, line 35: is this code extracted from some public library? if so I think it would be good to write a comment pointing to the maintainer of the code and the repo...
If not, it would be nice to add some tests as we now know 😄

khaledsherkawi: This is an npm package that also has browser support. I added all the details in the file.

Yelko-mrbeam (approved):

PR #1435, opened 2022-02-17
SW-1060 Inkscape version in SVG is incompatible with semver comparison library

khaledsherkawi:
Inkscape version embedded in SVGs are not Semantic and thus are incompatible with the semver library

### DIFF
--- a/octoprint_mrbeam/static/js/helpers/working_area_helper.js
+++ b/octoprint_mrbeam/static/js/helpers/working_area_helper.js
@@ -1,48 +1,4 @@
 class WorkingAreaHelper {
-    static versionCompare(v1, v2, options) {
-        var lexicographical = options && options.lexicographical,
-            zeroExtend = options && options.zeroExtend,
-            v1parts = v1.split("."),
-            v2parts = v2.split(".");
-
-        function isValidPart(x) {
-            return (lexicographical ? /^\d+[A-Za-z]*$/ : /^\d+$/).test(x);
-        }
-
-        if (!v1parts.every(isValidPart) || !v2parts.every(isValidPart)) {
-            return NaN;
-        }
-
-        if (zeroExtend) {
-            while (v1parts.length < v2parts.length) v1parts.push("0");
-            while (v2parts.length < v1parts.length) v2parts.push("0");
-        }
-
-        if (!lexicographical) {
-            v1parts = v1parts.map(Number);
-            v2parts = v2parts.map(Number);
-        }
-
-        for (var i = 0; i < v1parts.length; ++i) {
-            if (v2parts.length === i) {
-                return 1;
-            }
-
-            if (v1parts[i] === v2parts[i]) {
-            } else if (v1parts[i] > v2parts[i]) {
-                return 1;
-            } else {
-                return -1;
-            }
-        }
-
-        if (v1parts.length !== v2parts.length) {
-            return -1;
-        }
-
-        return 0;
-    }
-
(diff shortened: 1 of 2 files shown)


octoprint_mrbeam/static/js/helpers/working_area_helper.js
```diff
@@ -1,4 +1,48 @@
 class WorkingAreaHelper {
+    static versionCompare(v1, v2, options) {
+        var lexicographical = options && options.lexicographical,
```
Yelko-mrbeam: it would be nice to rename this to areLexicographicalVersions to know we are expecting to find a boolean here

octoprint_mrbeam/static/js/helpers/working_area_helper.js
```diff
@@ -1,4 +1,48 @@
...
+            zeroExtend = options && options.zeroExtend,
```
Yelko-mrbeam: same here... We could rename this to isStrictComparison. on the documentation we could then explain what strict means. which is 2.1 != 2.1.0 and not strict will do 2.1 == 2.1.0.

Yelko-mrbeam (commented):
Looks good! I added 2 suggestions but not really important since we do know this is old code that is being readded.
Since the difference with the comparator added on the previous pull request it's most likely just the addition of the validator (lexicographical) and the preformatter zeroExtend I would suggest to move this in to different methods and apply them before calling the new window.compareVersions.
What do you guys think? @ahmed-mrbeam @khaledsherkawi

octoprint_mrbeam/static/js/helpers/working_area_helper.js
```diff
@@ -1,4 +1,48 @@
...
+            v1parts = v1.split("."),
+            v2parts = v2.split(".");
+
+        function isValidPart(x) {
+            return (lexicographical ? /^\d+[A-Za-z]*$/ : /^\d+$/).test(x);
+        }
+
+        if (!v1parts.every(isValidPart) || !v2parts.every(isValidPart)) {
+            return NaN;
```
Yelko-mrbeam: I am also surprised this is not crashing the comparison... 🤔  NaN <= 0 returns false for some reason which escapes my understanding... Lovely JavaScript....

Yelko-mrbeam: also NaN >= 0 is false... In any case I think we shouldn't return 2 different types from one function... 😓

octoprint_mrbeam/static/js/helpers/working_area_helper.js
khaledsherkawi: we actually changed the whole handling. Will send you a new review request once done.

octoprint_mrbeam/static/js/helpers/working_area_helper.js
khaledsherkawi: we actually changed the whole handling. Will send you a new review request once done.

octoprint_mrbeam/static/js/helpers/working_area_helper.js
khaledsherkawi: Yea while Ahmed and I were going through this we found out that if the versions looks like: "1.1.2 (1:1.1+202202050950+0a00cf5339)" then this method will always return NaN which after applied to: NaN <= 0 it will always return false meaning all Inkscape designs passing through this has the new version scaling all the time.

Yelko-mrbeam (approved):

PR #1446, opened 2022-03-25
SW-1077-increase-laserhead-max-working-temperature

ahmed-mrbeam:
* Temperature manager to use the current Max Temp value for laser head during init procedure
* In case the laser head changes, The temperature_manager will read the new Max temp value based on the new Laser head detected
* Added new iobeam internal event in case of laserhead changed

### DIFF
--- a/octoprint_mrbeam/iobeam/temperature_manager.py
+++ b/octoprint_mrbeam/iobeam/temperature_manager.py
@@ -51,7 +51,7 @@ class TemperatureManager(object):
 
         self.dev_mode = plugin._settings.get_boolean(["dev", "iobeam_disable_warnings"])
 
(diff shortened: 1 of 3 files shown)


octoprint_mrbeam/iobeam/temperature_manager.py
```diff
@@ -70,14 +70,16 @@ def _on_mrbeam_plugin_initialized(self, event, payload):
         self._iobeam = self._plugin.iobeam
         self._analytics_handler = self._plugin.analytics_handler
         self._one_button_handler = self._plugin.onebutton_handler
-
+        self._subscribe()
```
ahmed-mrbeam, line 73: This change is not related to the fix
it is done because we shouldn't start the timer before all inits are done and the subscribe is and important step during init

JFSolution (approved):

PR #1447, opened 2022-03-30
Bugfix/sw 300 software update refresh issue

khaledsherkawi:


octoprint_mrbeam/templates/mrbeam_initscript.jinja2
```diff
@@ -1,5 +1,7 @@
 <script lang="javascript">
     var BEAMOS_VERSION = "{{ beamosVersionNumber }}";
+    // PLUGINS_VERSIONS is set by single quotes since "pluginVersions" is returning a JSON dump which will break with double quotes
+    var PLUGINS_VERSIONS = '{{ plugins_versions }}';
```
Yelko-mrbeam, line 4: Instead of using quotes you could try to use some of flask filters to make it easier. They seem to be pretty powerful.
Here is an example that would match this needs using tojson and safe https://stackoverflow.com/questions/24719592/sending-data-as-json-object-from-python-to-javascript-with-jinja
You would also need to remove the json.dumps call on the initialisation.

octoprint_mrbeam/util/device_info.py
```diff
@@ -137,6 +138,42 @@ def get_beamos_version_number(self):
             self._logger.debug("beamos_version invalid: " + beamos_version)
             return None
 
+    def get_software_versions(self):
+        result = dict()
+        software_info = None
+
+        try:
+            plugin_info = self._plugin._plugin_manager.get_plugin_info("softwareupdate")
```
Yelko-mrbeam, line 146: this is softwareupdate_plugin_info right?

This is also the only place where we need the plugin... I don't think we need to store a reference to the plugin from the DeviceInfo. It seems like to initialise this class we need to pass any plugin to it... Doesn't feel like it belongs here... If we need the softwareupdate info  I think we should directly pass it in the constructor.
Another flag that shows that this is smelly is that we are directly calling the _plugin_manager which is a private variable which means it shouldn't be used outside of its own class.

Yelko-mrbeam (changes requested):

PR #1459, opened 2022-04-13
SW-1208-mr-beam-plugin-will-break-if-cam-last-session-yaml-contains-null-chars

ahmed-mrbeam:
- Refactored the method `load_camera_settings` to handle a corrupt last_session.yaml file properly
- Added more exceptions to be handled
- created constants for used paths to files

### DIFF
--- a/octoprint_mrbeam/iobeam/lid_handler.py
+++ b/octoprint_mrbeam/iobeam/lid_handler.py
@@ -1358,12 +1358,11 @@ class PhotoCreator(object):
                 # No shutter speed info present in this file
                 settings = {k: v[-1] for k, v in settings.items()}
                 camera_settings = [settings, None]
-            self._logger.debug("lid_handler: Fallback camera settings loaded from file: {}".format(__file__, backup_path))
+            self._logger.debug("lid_handler: Fallback camera settings loaded from file: {}".format(backup_path))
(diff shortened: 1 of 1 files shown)


octoprint_mrbeam/iobeam/lid_handler.py
```diff
@@ -1371,10 +1391,10 @@ def save_camera_settings(
         settings = {}
         try:
             with open(path) as f:
-                settings = yaml.load(f)
-        except (OSError, IOError) as e:
+                settings = yaml.safe_load(f)
+        except (OSError, IOError, yaml.YAMLError, yaml.reader.ReaderError, AttributeError) as e:
             self._logger.warning(
```
khaledsherkawi, line 1396: This should be `self._logger.warn`
I'm not sure if it works with `warning` too 
I know it's not part of the changes but just wanted to mention
It's used in many locations in our plugin but in OctoPrint it is always: `self._logger.warn`

khaledsherkawi (approved):

octoprint_mrbeam/iobeam/lid_handler.py
ahmed-mrbeam: we have warning method in Mr beam plugin

PR #1463, opened 2022-04-21
SW-1223 don't show the update failed info message on boot

JFSolution:
add user clicked update flag
redirect softwareupdateplugin check for update to our endpoint if the user clicked on it
refactor fallback update config for offline

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -175,6 +175,7 @@ class MrBeamPlugin(
         self._serial_num = None
         self._mac_addrs = dict()
         self._model_id = None
+        self._user_checks_for_updates = False
         self._grbl_version = None
         self._device_series = self._device_info.get_series()
         self.called_hosts = []
@@ -737,6 +738,15 @@ class MrBeamPlugin(
         ret = check_calibration_tool_mode(self)
         self._fixEmptyUserManager()
         return ret
+
+    @property
+    def user_checks_for_updates(self):
+        return self._user_checks_for_updates
+
+    @user_checks_for_updates.setter
+    def user_checks_for_updates(self, value):
+        self._logger.debug("set plugin.user_checks_for_updates {}".format(value))
+        self._user_checks_for_updates = value
 
     ##~~ UiPlugin mixin
 
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -175,7 +175,7 @@ class MrBeamPlugin(
         self._serial_num = None
         self._mac_addrs = dict()
         self._model_id = None
-        self._user_checks_for_updates = False
+        self._explicit_update_check = False
         self._grbl_version = None
         self._device_series = self._device_info.get_series()
         self.called_hosts = []
@@ -740,13 +740,14 @@ class MrBeamPlugin(
         return ret
 
     @property
-    def user_checks_for_updates(self):
-        return self._user_checks_for_updates
+    def explicit_update_check(self):
+        return self._explicit_update_check
 
-    @user_checks_for_updates.setter
-    def user_checks_for_updates(self, value):
-        self._logger.debug("set plugin.user_checks_for_updates {}".format(value))
-        self._user_checks_for_updates = value
+    def set_explicit_update_check(self):
+        self._explicit_update_check = True
+
+    def clear_explicit_update_check(self):
+        self._explicit_update_check = False
 
     ##~~ UiPlugin mixin
 
--- a/octoprint_mrbeam/rest_handler/update_handler.py
+++ b/octoprint_mrbeam/rest_handler/update_handler.py
@@ -17,9 +17,6 @@ class UpdateRestHandlerMixin:
         if hasattr(request, "json") and request.json:
             data = request.json
             clicked_by_user = data.get("user", False)
-        self._logger.debug(
-            "update rest handler data {} {}".format(clicked_by_user, data)
-        )
         reload_update_info(self, clicked_by_user)
         return self._plugin_manager.get_plugin_info(
             "softwareupdate"
--- a/octoprint_mrbeam/rest_handler/update_handler.py
+++ b/octoprint_mrbeam/rest_handler/update_handler.py
@@ -12,7 +12,6 @@ class UpdateRestHandlerMixin:
 
     @octoprint.plugin.BlueprintPlugin.route("/info/update", methods=["POST"])
     def update_update_informations(self):
-        data = None
         clicked_by_user = False
         if hasattr(request, "json") and request.json:
             data = request.json
--- a/octoprint_mrbeam/rest_handler/update_handler.py
+++ b/octoprint_mrbeam/rest_handler/update_handler.py
@@ -1,4 +1,4 @@
-from octoprint.server import NO_CONTENT
+from flask import request
 
 import octoprint.plugin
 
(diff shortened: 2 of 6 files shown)


octoprint_mrbeam/__init__.py
```diff
@@ -738,6 +739,15 @@ def calibration_tool_mode(self):
         self._fixEmptyUserManager()
         return ret
 
+    @property
+    def user_checks_for_updates(self):
```
ahmed-mrbeam: suggest the following refactor:
`user_checks_for_updates` --> `explicit_Update_Check`

octoprint_mrbeam/__init__.py
```diff
@@ -738,6 +739,15 @@ def calibration_tool_mode(self):
...
+        return self._user_checks_for_updates
+
+    @user_checks_for_updates.setter
+    def user_checks_for_updates(self, value):
```
ahmed-mrbeam: Since it is a boolean value, I would suggest to create two setters, one for set and one for clear

octoprint_mrbeam/rest_handler/update_handler.py
```diff
@@ -12,5 +12,12 @@ class UpdateRestHandlerMixin:
 
     @octoprint.plugin.BlueprintPlugin.route("/info/update", methods=["POST"])
     def update_update_informations(self):
-        reload_update_info(self)
-        return NO_CONTENT
+        data = None
+        clicked_by_user = False
+        if hasattr(request, "json") and request.json:
+            data = request.json
+            clicked_by_user = data.get("user", False)
```
ahmed-mrbeam, line 18: Are you sure you want to default to False ?

octoprint_mrbeam/software_update_information.py
```diff
@@ -54,6 +54,30 @@ class SWUpdateTier(Enum):
 BEAMOS_LEGACY_VERSION = "0.14.0"
 BEAMOS_LEGACY_DATE = date(2018, 1, 12)  # still used in the migrations
 
+FALLBACK_UPDATE_CONFIG = {
```
ahmed-mrbeam, line 57: Is this change related to the fix ?

octoprint_mrbeam/software_update_information.py
```diff
@@ -168,32 +193,7 @@ def get_update_information(plugin):
...
-                "findmymrbeam": {
-                    "name": "OctoPrint-FindMyMrBeam",
-                    "type": "github_commit",
-                    "user": "",
-                    "repo": "",
-                    "pip": "",
-                },
-            },
-        },
+        {},
```
ahmed-mrbeam, line 196: I don't understand what happened here

octoprint_mrbeam/software_update_information.py
```diff
@@ -229,14 +229,15 @@ def switch_software_channel(plugin, channel):
     Returns:
         None
     """
+    _logger.debug("switch_software_channel")
```
ahmed-mrbeam, line 232: Could you please add at the begging of the log something  like "Software update information: "
so that it can be clear in the logs?

octoprint_mrbeam/software_update_information.py
```diff
@@ -229,14 +229,15 @@ def switch_software_channel(plugin, channel):
...
     old_channel = plugin._settings.get(["dev", "software_tier"])
     if channel in software_channels_available(plugin) and channel != old_channel:
         _logger.info("Switching software channel to: {channel}".format(channel=channel))
         plugin._settings.set(["dev", "software_tier"], channel)
         reload_update_info(plugin)
 
 
-def reload_update_info(plugin):
+def reload_update_info(plugin, clicked_by_user=False):
```
ahmed-mrbeam, line 240: Not sure why we have `clicked_by_user` and _user_checks_for_updates why not use one of them ?

ahmed-mrbeam (commented):
Partially reviewed

octoprint_mrbeam/rest_handler/update_handler.py
JFSolution: Yes, if it is not set the default should be false, as we only set it to true if the users clicks on it and otherwise leave this data element empty

octoprint_mrbeam/software_update_information.py
JFSolution: this will be added from the logger it selfe as it logs everything from this file as `octoprint.plugins.mrbeam.software_update_information`

octoprint_mrbeam/software_update_information.py
JFSolution: Yes I needed to refactor this so I am still able to give back the fallback update config

octoprint_mrbeam/software_update_information.py
JFSolution: this is releated to the refactoring https://github.com/mrbeam/MrBeamPlugin/pull/1463/files#r859892304
I refactored it so the _set_info_from_cloud_config method will return the `FALLBACK_UPDATE_CONFIG` if something is not correct instead of returning it only if there was a connection problem

octoprint_mrbeam/software_update_information.py
JFSolution: clicked by user is only the input parameter of reload_update_info, that will then be saved to the user_checks_for_updates

octoprint_mrbeam/__init__.py
JFSolution: ok i can change that

octoprint_mrbeam/software_update_information.py
```diff
@@ -245,6 +246,9 @@ def reload_update_info(plugin):
     Returns:
         None
     """
+    if clicked_by_user:
+        plugin.user_checks_for_updates = True
```
ahmed-mrbeam: please use the setter

octoprint_mrbeam/software_update_information.py
```diff
@@ -320,17 +325,20 @@ def _set_info_from_cloud_config(plugin, tier, beamos_version, cloud_config):
                 f.write(json.dumps(sw_update_config))
         except (IOError, TypeError):
             plugin._logger.error("can't create update info file")
```
ahmed-mrbeam: Please add here that Update fallback config will be used

octoprint_mrbeam/static/js/software_channel_selector.js
```diff
@@ -95,11 +95,12 @@ $(function () {
         // get the hook when softwareUpdate perform the Updatecheck to force the update on the normal button
         self.performCheck_copy = self.softwareUpdate.performCheck;
         self.softwareUpdate.performCheck= function(showIfNothingNew, force, ignoreSeen) {
-            self.reload_update_info();
             if (force !== undefined) {
                 force = true; //only forces the update check if it was disabled ("check for update" button press)
+                self.reload_update_info(showIfNothingNew, force, ignoreSeen, true); //if use clicked update forward to our reload endpoint
```
ahmed-mrbeam, line 100: `showIfNothingNew` `ignoreSeen`   why using different naming convention here ?

tests/softwareupdate/test_cloud_config.py
```diff
@@ -461,10 +438,11 @@ def test_cloud_confg_fileerror(
                 },
             )
             plugin = self.plugin
+            plugin.user_checks_for_updates = True
```
ahmed-mrbeam: Please use the setter

octoprint_mrbeam/__init__.py
JFSolution: I can do this

octoprint_mrbeam/__init__.py
JFSolution: will change it to set_explicit_update_check and clear_explicit_update_check

ahmed-mrbeam (approved):

ahmed-mrbeam (approved):

PR #1471, opened 2022-04-27
SW-1224 fix the legacy update problem

JFSolution:


JFSolution:
wuups these are a lot of merge conflicts

JFSolution:
maybe we need to force push

PR #1483, opened 2022-05-05
SW-1262 add material setting for marker spray

JFSolution:
- adding the marker spray material settings
- added translation for material setting
- version bump


JFSolution:
The PR is directly to stable so please don't merge now

khaledsherkawi (approved):

PR #1487, opened 2022-05-16
SW-1291 add Dockerfile

Yelko-mrbeam:
Dockerfile is created.
you can run the image by executing
`docker build -t mrbeam_plugin .`
and then
`docker run --name mrbeam-plugin -d -p5002:5000 mrbeam_plugin`
You'll be able to access the instance under
`localhost:5002`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,8 +36,12 @@ nbproject
 .directory
 plugvenv/
 venv/
+
 # Pytest files
 tests/rsc/camera/debug/[^R]*
 tests/rsc/camera/out.jpg
 tests/logs/*.txt
 *pytest-logs.txt
+
+# node modules
+node_modules
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.10
+FROM ubuntu:21.10
 
 ENV TZ=Europe/Berlin
 # Installing python and pip on user level so we need to add it to the PATH
@@ -59,7 +59,7 @@ COPY . /home/mrbeam/MrBeamPlugin/
 RUN source ./venv2/bin/activate \
     && sudo -u mrbeam env "PATH=$PATH" pip install ./MrBeamPlugin
 
-CMD [ "./venv2/bin/python", "-m", "octoprint", "serve", "--debug", "--port", "5000" ]
+CMD [ "./venv2/bin/python", "-m", "octoprint", "serve", "--port", "5000" ]
 
 HEALTHCHECK  --interval=30s --timeout=3s \
     CMD wget --no-verbose --tries=3 --spider http://localhost:5000 || exit 1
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -36,8 +36,12 @@ nbproject
...
 venv/
+
 # Pytest files
 tests/rsc/camera/debug/[^R]*
 tests/rsc/camera/out.jpg
 tests/logs/*.txt
 *pytest-logs.txt
+
+# node modules
+node_modules
```
JFSolution, line 47: why we need node_modules?

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+ENV TZ=Europe/Berlin
+# Installing python and pip on user level so we need to add it to the PATH
+ENV PATH="/home/mrbeam/.local/bin:/home/mrbeam/bin:${PATH}"
+
+SHELL ["/bin/bash", "-c"]
+
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+RUN mkdir /home/mrbeam/ \
+    && groupadd -g 999 mrbeam \
```
JFSolution, line 12: Should we change the user to pi, so it will be the same as on our image?

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+
+RUN source ./venv2/bin/activate \
+    && pip install opencv-python==4.2.0.32
+
+COPY . /home/mrbeam/MrBeamPlugin/
+
+RUN source ./venv2/bin/activate \
+    && sudo -u mrbeam env "PATH=$PATH" pip install ./MrBeamPlugin
+
+CMD [ "./venv2/bin/python", "-m", "octoprint", "serve", "--debug", "--port", "5000" ]
```
JFSolution: I think we shouldn't use the debug flag here, as we should test against the same scenario like it will run for the user

.gitignore
Yelko-mrbeam: This will appear once we add javascript tests and we shouldn't upload this to the repo

Dockerfile
Yelko-mrbeam: Hmm... Good question... Why is it called pi? If we call it "pi" here it will be confusing, I think. This is meant to be run anywhere not in a raspberry... 🤔

Dockerfile
Yelko-mrbeam: you are right!

Dockerfile
JFSolution: It will work with mrbeam too, was just an idea to have the same user. But it is fine for me with the mrbeam user too

.gitignore
JFSolution: But this is not related to the docker image and you are already merging into alpha

JFSolution (approved):

PR #1491, opened 2022-05-18
SW-1269 usage file and backup file will break if version number is not text only

JFSolution:
- repair broken backup file
- convert unicode version number to str

### DIFF
--- a/octoprint_mrbeam/analytics/usage_handler.py
+++ b/octoprint_mrbeam/analytics/usage_handler.py
@@ -1,5 +1,7 @@
 import os
 import time
+import unicodedata
+
 import yaml
 
 from octoprint_mrbeam.mrb_logger import mrb_logger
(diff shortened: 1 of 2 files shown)


octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -356,36 +358,59 @@ def _load_usage_data(self):
...
-            if os.path.isfile(self._backup_file):
+            try:
+                with open(self._backup_file, "r") as stream:
+                    data = yaml.safe_load(stream)
+                if self._validate_data(data):
+                    data["restored"] = (
+                        data["restored"] + 1 if "restored" in data else 1
+                    )
+                    self._usage_data = data
+                    success = True
```
ahmed-mrbeam: is it safe to set this variable to True before writing to the file ?

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -356,36 +358,59 @@ def _load_usage_data(self):
...
+            except:
+                self._logger.error("Can't read _backup_file file.")
 
         if not success:
             self._logger.warn("Resetting usage data. (marking as incomplete)")
             self._usage_data = self._get_usage_data_template()
             if recovery_try:
                 self._write_usage_data()
 
+    def _repair_backup_usage_data(self):
```
ahmed-mrbeam, line 390: Please add method doc string

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -356,36 +358,59 @@ def _load_usage_data(self):
...
+        success = False
+        with open(self._backup_file, "r") as stream:
+            data = yaml.load(stream)
+        if self._validate_data(data):
+            if isinstance(data["version"], unicode):
```
ahmed-mrbeam, line 402: Please add a comment explaining why we need this and put the link to the Jira issue in the comment as well

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -356,36 +358,59 @@ def _load_usage_data(self):
...
+                data["version"] = unicodedata.normalize('NFKD', data["version"]).encode('ascii', 'ignore')
+            data["restored"] = (
+                data["restored"] + 1 if "restored" in data else 1
```
ahmed-mrbeam, line 405: what is that line supposed to do ?

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: I can change the order

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: ok, will do

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: ok

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: the restored? It will count how often this file was restored from the backup file

Yelko-mrbeam (approved):

PR #1501, opened 2022-05-29
Feature/sw 1313 add sonar cloud

ahmed-mrbeam:


sonar-project.properties
```diff
@@ -0,0 +1,14 @@
+sonar.projectKey=mrbeam_MrBeamPlugin
+sonar.organization=mrbeam
+
+# This is the name and version displayed in the SonarCloud UI.
+sonar.projectName=MrBeamPlugin
+sonar.projectVersion=v0.11.0b3
```
ahmed-mrbeam, line 6: Not sure about this one. May be it needs to be revised later.
Here I just put the latest release

JFSolution: I can't see the version in the sonar cloud UI

JFSolution (approved):

sonar-project.properties
ahmed-mrbeam: me neither . needs to be investigated more

PR #1503, opened 2022-05-31
SW-1278 quicktext fill color is not being engraved in the latest svg feature branch

khaledsherkawi:
- Modify filled vector detection method for quicktext elements
- Control the filling inclusion/exclusion of the quicktext elements with fill-opacity
- Add white stroke color for stroke exclusion and white filling when none exists for easier cursor selection


JFSolution (approved):
Khaled walked me trough the PR and everything looks good

PR #1504, opened 2022-05-31
SW-1317 add pressure sensor data to analytics

JFSolution:
- added pressure sensor data to analytics

### DIFF
--- a/octoprint_mrbeam/iobeam/dust_manager.py
+++ b/octoprint_mrbeam/iobeam/dust_manager.py
@@ -145,6 +145,15 @@ class DustManager(object):
         self._event_bus.subscribe(OctoPrintEvents.SHUTDOWN, self._onEvent)
 
     def _handle_exhaust_data(self, args):
+        """
+        hanldes exhaust data comming from iobeam EXHAUST_DYNAMIC_VALUE event
+
+        Args:
+            args: data from the iobeam event
+
+        Returns:
+            None
+        """
         self._logger.debug("last pressure values append {} - {}".format(args["pressure"], self._last_pressure_values))
         if args["pressure"] is not None:
             self._last_pressure_values.append(args["pressure"])
--- a/octoprint_mrbeam/iobeam/dust_manager.py
+++ b/octoprint_mrbeam/iobeam/dust_manager.py
@@ -270,28 +270,13 @@ class DustManager(object):
         try:
             # Write to analytics if the values are valid
             if self._validate_values():
-                if len(self._last_rpm_values):
-                    rpm_average = sum(self._last_rpm_values) / len(
-                        self._last_rpm_values
-                    )
-                else:
-                    rpm_average = -1
-
-                if len(self._last_pressure_values):
-                    pressure_average = sum(self._last_pressure_values) / len(
-                        self._last_pressure_values
-                    )
-                else:
-                    pressure_average = -1
-                self._logger.debug("pressure values mean {}".format(pressure_average))
-
                 data = dict(
-                    rpm_val=rpm_average,
+                    rpm_val=list(self._last_rpm_values),
                     fan_state=self._state,
                     usage_count=self._usage_handler.get_total_usage(),
                     prefilter_count=self._usage_handler.get_prefilter_usage(),
                     carbon_filter_count=self._usage_handler.get_carbon_filter_usage(),
-                    pressure_val=pressure_average,
+                    pressure_val=list(self._last_pressure_values),
                 )
                 self._analytics_handler.add_fan_rpm_test(data)
 
--- a/octoprint_mrbeam/iobeam/dust_manager.py
+++ b/octoprint_mrbeam/iobeam/dust_manager.py
@@ -121,7 +121,9 @@ class DustManager(object):
 
     def _subscribe(self):
         self._iobeam.subscribe(IoBeamValueEvents.DYNAMIC_VALUE, self._handle_fan_data)
-        self._iobeam.subscribe(IoBeamValueEvents.EXHAUST_DYNAMIC_VALUE, self._handle_exhaust_data)
+        self._iobeam.subscribe(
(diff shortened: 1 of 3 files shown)


octoprint_mrbeam/iobeam/dust_manager.py
```diff
@@ -142,6 +144,21 @@ def _subscribe(self):
...
 
+    def _handle_exhaust_data(self, args):
+        """
+        hanldes exhaust data comming from iobeam EXHAUST_DYNAMIC_VALUE event
+
+        Args:
+            args: data from the iobeam event
+
+        Returns:
+            None
```
ahmed-mrbeam: I would leave it empty.. as `None` is a keyword

octoprint_mrbeam/iobeam/dust_manager.py
```diff
@@ -142,6 +144,21 @@ def _subscribe(self):
...
+        """
+        self._logger.debug("last pressure values append {} - {}".format(args["pressure"], self._last_pressure_values))
+        if args["pressure"] is not None:
```
ahmed-mrbeam: what id` "pressure" `is not in args ?

octoprint_mrbeam/iobeam/dust_manager.py
JFSolution: ok will change

octoprint_mrbeam/iobeam/dust_manager.py
JFSolution: good catch, will change it

octoprint_mrbeam/iobeam/iobeam_handler.py
```diff
@@ -1068,7 +1069,13 @@ def _handle_exhaust(self, dataset):
         :param dataset:
         :return: error count
         """
-        # self._logger.info("exhaust dataset: '%s'", dataset)
+        self._logger.debug("exhaust dataset: '%s'", dataset)
+        # get the pressure sensor reading this will come as dust with the current iobeam version
+        if "dust" in dataset:
+            vals = {
```
ahmed-mrbeam, line 1075: not a good name

octoprint_mrbeam/iobeam/dust_manager.py
```diff
@@ -254,12 +271,19 @@ def _finish_test_fan_rpm(self):
                 else:
                     rpm_average = -1
 
+                if len(self._last_pressure_values):
+                    pressure_average = sum(self._last_pressure_values) /len(self._last_pressure_values)
```
ahmed-mrbeam: I'm curious , why we need to calculate the average and then log to analytics. wont we benefit more of the data if we log all of it and then on the cloud part we have `Raw Data` that we can see and analyse the way we see fit ?

JFSolution: It will be more data, but you are right, we can analyse it better with more data

octoprint_mrbeam/iobeam/iobeam_handler.py
JFSolution: for what?

octoprint_mrbeam/iobeam/dust_manager.py
JFSolution: The analytics entry will look now like this
`{"sw_tier": "DEV", "ntp": false, "sid": "94240d6fba2b41a19a1a6f42c8a7ca0e@00000000B135A71F-2F", "version_mrbeam_plugin": "0+unknown", "snr": "00000000B135A71F-2F", "data": {"rpm_val": [4450.0, 4617.0, 4737.0, 4795.0, 4881.0], "usage_count": 6164.9860281944275, "carbon_filter_count": 8029.77774746418, "fan_state": 50.0, "pressure_val": [1000, 1000, 1000, 1000, 1000], "prefilter_count": 8029.77774746418}, "uptime": 18410.01, "e": "fan_rpm_test", "job_id": "j_00000000B135A71F-2F_1654174137.3", "ts": 1654174177.719735, "t": "job", "env": "DEV", "v": 23, "model": "MRBEAM2_DC_S"}`

octoprint_mrbeam/iobeam/dust_manager.py
```diff
@@ -247,19 +270,13 @@ def _finish_test_fan_rpm(self):
...
-                if len(self._last_rpm_values):
-                    rpm_average = sum(self._last_rpm_values) / len(
-                        self._last_rpm_values
-                    )
-                else:
-                    rpm_average = -1
-
                 data = dict(
-                    rpm_val=rpm_average,
+                    rpm_val=list(self._last_rpm_values),
```
Yelko-mrbeam, line 274: We are pushing a list to analytics? is that ok in order to query it afterwards or how are we going to check this data afterwards? @Josef-MrBeam

Yelko-mrbeam, line 274: We are pushing a list to analytics? is that ok in order to query it afterwards or how are we going to check this data afterwards? @Josef-MrBeam

octoprint_mrbeam/iobeam/iobeam_handler.py
Yelko-mrbeam: I guess he was referring to vals. we could call it pressure_values 🤔

Yelko-mrbeam: I guess he was referring to vals. we could call it pressure_values 🤔

octoprint_mrbeam/iobeam/dust_manager.py
JFSolution: Ahmed said it would make more sense to push the whole list instead of a mean value, so we can calculate these during the analysis off the data.

Yelko-mrbeam: ok, I guess there is some thought put into how to deal with the data afterwards 👍

Yelko-mrbeam (approved):

PR #1507, opened 2022-06-09
Sw 1290  avoid crashing on import failure

Yelko-mrbeam:
https://mr-beam.atlassian.net/browse/SW-1290

### DIFF
--- a/octoprint_mrbeam/model/empty_import.py
+++ b/octoprint_mrbeam/model/empty_import.py
@@ -9,7 +9,7 @@ class EmptyImport(object):
     def log_error(self):
         try:
             from octoprint_mrbeam import mrb_logger
-            logger = mrb_logger("octoprint.plugins.mrbeam")
+            logger = mrb_logger("octoprint.plugins.mrbeam.model.EmptyImport")
             logger.error(self.error_message)
         except ImportError:
             return False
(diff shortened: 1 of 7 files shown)


octoprint_mrbeam/model/empty_import.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, module, default_return=None):
+        super(EmptyImport, self).__init__()
+        self.default_return = default_return
+        self.error_message = 'Error occurred trying to import module {module}'.format(module=module)
+        self.log_error()
+
+    def log_error(self):
+        try:
+            from octoprint_mrbeam import mrb_logger
+            logger = mrb_logger("octoprint.plugins.mrbeam")
```
JFSolution: this should include a better description of the module that logs, just mrbeam is to wide

JFSolution: maybe something like octoprint.plugins.mrbeam.model.empty_import, or you coudl use the filename

JFSolution:
Will the user see any error or info that docs is not working?

octoprint_mrbeam/model/empty_import.py
Yelko-mrbeam: Nice catch 👏

Yelko-mrbeam: I added the string we would need to add on the import in order to use this class we could stick with this rule and actually maybe create a decorator to inject a logger to a class using this rule 🤔  shall we create a techdebt ticket for this? maybe @irlaec wants to take this as a first challenge 😋

JFSolution (approved):

PR #1508, opened 2022-06-10
SW-1588 Cypress: Purchase designs - e2e-tests

Agnieszka12ax:
E2E-tests for purchase designs. 
How run test: npm run cypress:open


package.json
```diff
@@ -0,0 +1,6 @@
+{
```
ahmed-mrbeam, line 1: Would suggest to add the following:
`"scripts": {
    "cypress:open": "cypress open"
  }`
  
  Reference:
  https://docs.cypress.io/guides/getting-started/opening-the-app#cypress-open

package.json
```diff
@@ -0,0 +1,16 @@
...
+  "devDependencies": {
+    "@4tw/cypress-drag-drop": "^2.2.1",
+    "cypress": "^10.4.0",
+    "cypress-file-upload": "^5.0.8",
+    "cypress-iframe": "^1.0.1",
+    "prettier": "^2.7.1"
```
irlaec: Do you think we can/should remove prettier from here, as we already have it in pre-commit? Right now we have the same version in both, but in case we change it

irlaec (approved):

PR #1513, opened 2022-06-28
SW-1265-Try-to-run-Mr-Beam-plugin-with-Octoprint-1.7.3

JFSolution:
fixed only the errors to get the plugin running under oprint 1.7.3

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -762,6 +762,8 @@ class MrBeamPlugin(
         # template, using the render_kwargs as provided by OctoPrint
         from flask import make_response, render_template, g
 
+        sockjs_connect_timeout = settings().getInt(["devel", "sockJsConnectTimeout"])
+
         firstRun = render_kwargs["firstRun"]
         language = g.locale.language if g.locale else "en"
 
@@ -815,6 +817,7 @@ class MrBeamPlugin(
                 safetyGlasses=safety_glasses,
                 enableTemperatureGraph=False,
                 enableAccessControl=enable_accesscontrol,
+                sockJsConnectTimeout=sockjs_connect_timeout * 1000,
                 accessControlActive=accesscontrol_active,
                 enableSdSupport=False,
                 gcodeMobileThreshold=0,
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -18,7 +18,7 @@ from subprocess import check_output
 import octoprint.plugin
 import requests
 from flask import request, jsonify, make_response, url_for
-from flask.ext.babel import gettext
+from flask_babel import gettext
 import octoprint.filemanager as op_filemanager
 from octoprint.filemanager import ContentTypeDetector, ContentTypeMapping, FileManager
 from octoprint.server import NO_CONTENT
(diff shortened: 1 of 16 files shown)


octoprint_mrbeam/static/js/calibration/corner_calibration.js
```diff
@@ -426,7 +426,7 @@ $(function () {
         // USER CLICKS
         self.userClick = function (vm, ev) {
             // check if picture is loaded
-            if (window.location.href.indexOf("localhost") === -1)
+            if (window.location.getAttribute("href").indexOf("localhost") === -1)
```
Yelko-mrbeam: This seems quite fishy... Also why changing href for getAttribute? I think both would break in this case 🤔 
If we are just checking if we are running on the local server, we could do this instead:
`if (location.hostname === "localhost" || location.hostname === "127.0.0.1")`

octoprint_mrbeam/__init__.py
```diff
@@ -762,6 +762,8 @@ def on_ui_render(self, now, request, render_kwargs):
         # template, using the render_kwargs as provided by OctoPrint
         from flask import make_response, render_template, g
 
+        sockjs_connect_timeout = settings().getInt(["devel", "sockJsConnectTimeout"])
```
Yelko-mrbeam, line 765: what is devel?

JFSolution: devel is on off the octoprint config settings

octoprint_mrbeam/static/js/calibration/corner_calibration.js
Yelko-mrbeam: ah actually we are checking for the contrary 🤔 
So we are looking for localhost and if the index is -1 (Not found) then we enter the if...
We should extract this logic to a method but it would be nice to make it more readable like:

```
if(isNotLocalhost(location.hostname)) {
  ...
}

function isNotLocalhost(hostname) {
  return !isLocalhost(hostname);
}

function isLocalhost(hostname) {
  return hostname === "localhost" || isHomeIp(hostname);
}

function isHomeIP(ip) {
   let ipNumbers = ip.split('.');
   let isIpNumberLength = ipNumbers.length === 4;
   let isArrayOfNumbers = ipNumbers.every(element => {
     return !isNaN(element);
    });
   return isIpNumberLength && isArrayOfNumbers && ipNumbers[0] === '127';
}
```

JFSolution: good catch, but I tried with the hostname and it was not working. I ned to revert this cahnge.
But I will try to understand why this is even needed

octoprint_mrbeam/__init__.py
Yelko-mrbeam: what does it represent? it seems to me that it comes from development? which should mean it is only used for development? if we are using it on the released version I think it should be called something else... or am I understanding this wrong?

JFSolution: As far as I understand this can be changed for development purpose. But the default setting is 30.
This change is only to get this setting to the frontend otherwise the loading can't be finished as the timeout will hit before

Yelko-mrbeam: ok but I understand it shouldn't be called devel then right? no need to do it here but we could create a tech debt for it or a meeting to discuss it :)

Yelko-mrbeam (approved):

octoprint_mrbeam/__init__.py
JFSolution: This is a octoprint thing, and we are just using it, so it doesn't make sense to rename it

PR #1520, opened 2022-06-30
Bugfix/sw 1427 fix mr beam plugin unit tests

irlaec:
Most of the tests are now passing, except the ones related to Software Update.
The "frontend" tests were marked as deprecated, as we now do them in Cypress.


requirements.txt
```diff
@@ -1,9 +0,0 @@
-###
```
JFSolution: please keep the original file

JFSolution (changes requested):

requirements.txt
irlaec: Ooops, restored!

tests/frontend/critical_designs/base_procedure.py
```diff
@@ -1,17 +1,17 @@
...
 from selenium.webdriver.support import expected_conditions as EC
 
-from frontend import uiUtils
-from frontend import webdriverUtils
-from frontend import gcodeUtils
-from frontend import frontendTestUtils
+from .. import uiUtils
+from .. import webdriverUtils
+from .. import gcodeUtils
+from .. import frontendTestUtils
```
JFSolution: I don't like these .. does this work too?

```suggestion
from tests.frontend import webdriverUtils, gcodeUtils, frontendTestUtils, uiUtils
```

JFSolution (changes requested):
please change the from imports to absolute imports

tests/frontend/critical_designs/base_procedure.py
irlaec: I changed all the relative imports to absolute

JFSolution (approved):

PR #1521, opened 2022-06-30
SW-1430 enable laser button on dev devices by parameter

Yelko-mrbeam:
https://mr-beam.atlassian.net/browse/SW-1430

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -855,6 +855,7 @@ class MrBeamPlugin(
                     self.get_model_id()),
                 burger_menu_model=BurgerMenuService(self._logger, DocumentService(self._logger)).get_burger_menu_model(
                     self.get_model_id()),
+                isDevelop=self.is_dev_env(),
             )
         )
         r = make_response(render_template("mrbeam_ui_index.jinja2", **render_kwargs))
--- a/octoprint_mrbeam/templates/tab_workingarea.jinja2
+++ b/octoprint_mrbeam/templates/tab_workingarea.jinja2
@@ -50,7 +50,7 @@
 
 													<div class="row-fluid print-control btn-group" style="display: none;" data-bind="visible: loginState.isUser">
 														<button id="laser_button" class="btn btn-danger span6"
-                                                                data-bind="click: conversion.show_conversion_dialog, enable: state.isOperational() && state.isReady() && !state.isPrinting() && !state.isPaused() && loginState.isUser() && !workingArea.working_area_empty(), attr: {title: state.titlePrintButton}"
+                                                                data-bind="click: conversion.show_conversion_dialog, enable: {{ 'true' if isDevelop and 'enableLaserButton' in request.args else 'false'}} || (state.isOperational() && state.isReady() && !state.isPrinting() && !state.isPaused() && loginState.isUser() && !workingArea.working_area_empty()), attr: {title: state.titlePrintButton}"
                                                                 id="job_print">
                                                             <i class="icon-white icon-fire"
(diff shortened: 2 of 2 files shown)


octoprint_mrbeam/__init__.py
```diff
@@ -855,6 +855,7 @@ def on_ui_render(self, now, request, render_kwargs):
                     self.get_model_id()),
                 burger_menu_model=BurgerMenuService(self._logger, DocumentService(self._logger)).get_burger_menu_model(
                     self.get_model_id()),
+                isDevelop=self.is_dev_env(),
```
JFSolution, line 858: can this easily be changed by the user? So you 'hack' your device to be a develop one

octoprint_mrbeam/templates/tab_workingarea.jinja2
```diff
@@ -50,7 +50,7 @@
 
 													<div class="row-fluid print-control btn-group" style="display: none;" data-bind="visible: loginState.isUser">
 														<button id="laser_button" class="btn btn-danger span6"
-                                                                data-bind="click: conversion.show_conversion_dialog, enable: state.isOperational() && state.isReady() && !state.isPrinting() && !state.isPaused() && loginState.isUser() && !workingArea.working_area_empty(), attr: {title: state.titlePrintButton}"
+                                                                data-bind="click: conversion.show_conversion_dialog, enable: {{ 'true' if isDevelop and 'enableLaserButton' in request.args else 'false'}} || (state.isOperational() && state.isReady() && !state.isPrinting() && !state.isPaused() && loginState.isUser() && !workingArea.working_area_empty()), attr: {title: state.titlePrintButton}"
```
JFSolution, line 53: why you add this additional 'true' and 'false
it would be easier with
```suggestion
                                                                data-bind="click: conversion.show_conversion_dialog, enable: (isDevelop && 'enableLaserButton' in request.args) || (state.isOperational() && state.isReady() && !state.isPrinting() && !state.isPaused() && loginState.isUser() && !workingArea.working_area_empty()), attr: {title: state.titlePrintButton}"
```

JFSolution (changes requested):

octoprint_mrbeam/templates/tab_workingarea.jinja2
Yelko-mrbeam: Notice that the code I added is python code executed on the jinja2 templating engine while the rest of the condition is executed on the browser of the user. I need to echo either 'true' or 'false' as it would be written in javascript since also python would capitalise the boolean 🤦

octoprint_mrbeam/__init__.py
Yelko-mrbeam: I am not sure about this... But I did not change that behaviour, I am only explicitly exposing it to the jinja2 templating engine...

octoprint_mrbeam/templates/tab_workingarea.jinja2
JFSolution: Oh I didn't know that is possible 😅. This is needed to access the args right?
But do we still need to expose the isDevelop?

Yelko-mrbeam: this is only exposed to the jinja2 not to the frontend so it shouldn't be a problem... The end html will only contain either a true or a false there...

Yelko-mrbeam: we can have a quick sync tomorrow and we discuss further faster 😄

JFSolution (approved):

PR #1522, opened 2022-07-04
SW-1180 split the updatescript to use params from a config file

irlaec:
Move the params of the update script to a separate file, in order to use the same update script in different plugins


JFSolution (approved):

JFSolution (changes requested):
we need to remove the octoprint_mrbeam part completly from the update script, so it can be a clone in in the netconnectd plugin

JFSolution (changes requested):
please fix the reformatting, as this is not the same with black, but the rest is looking good

octoprint_mrbeam/scripts/update_script.py
```diff
@@ -404,30 +427,48 @@ def loadPluginTarget(archive, folder):
...
+        folder=folder,
+        plugin_extracted_path_folder=plugin_extracted_path_folder,
+        origin_file="scripts/update_script_config.py",
+        destination_file="update_script_config.py",
+    )
+
+    return zip_file_path
+
+
+def _copy_file_to_working_dir(
```
JFSolution, line 456: I like, that you split this in a new function, but please add a Google Doc string

irlaec: It's added :)

irlaec: It's added

JFSolution (approved):

PR #1523, opened 2022-07-05
SW-1328 the text of the svgs will not be engraved

khaledsherkawi:
- Hide fast engraving advanced option as Teja decided this option is too buggy and removed the option from the tooltip but forgot the actual button
- Fix MRBEAM_DEBUG_RENDERING debugging as the old method that was still used there breaks the debugging too
- Mark text elements as filled instead of tspan elements: Tspan elements cannot be rendered alone without a text parent leading to some SVG text combinations being skipped
- Add SVG styling to all clusters: Font Declarations are added for quickText fonts but the SVG styling has been left out leading to any text styling being missing. SVG styling is added the same way we are adding the font declarations into the different clusters.

### DIFF
--- a/octoprint_mrbeam/static/js/render_fills.js
+++ b/octoprint_mrbeam/static/js/render_fills.js
@@ -64,7 +64,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
                 className,
                 fillPaths
             );
-            if (Array.isArray(processedElem) && processedElem?.length === 0) {
+            if (Array.isArray(processedElem) && processedElem.length === 0) {
                 return [];
             } else {
                 selection.push(processedElem);
--- a/octoprint_mrbeam/static/js/render_fills.js
+++ b/octoprint_mrbeam/static/js/render_fills.js
@@ -39,6 +39,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
         // TODO: better than checking children.length and a blacklist would be to refer to
         // Graphic Elements vs. Container Elements (see https://www.w3.org/TR/SVG/struct.html#TermContainerElement)
         // image is excluded here as <image ...>\n</image> has a child of type #text (== '\n')
+        // text is excluded here as <tspan> cannot be rendered without a <text> parent
         const goRecursive = ![
             "image",
             "defs",
@@ -47,6 +48,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
             "rdf:rdf",
             "cc:work",
             "sodipodi:namedview",
+            "text",
         ].includes(elem.type);
 
         if (children.length > 0 && goRecursive) {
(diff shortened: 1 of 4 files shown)


octoprint_mrbeam/static/js/render_fills.js
```diff
@@ -64,6 +66,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
                 elem.type === "text" ||
                 elem.type === "#text"
             ) {
```
Yelko-mrbeam: This if condition is useless right now and only adds cognitive complexity... We could remove it.
It might also be good to move all the else branch corresponding to the `if (children.length > 0 && goRecursive) {` to a separate method which could be called processElementByType(type)

octoprint_mrbeam/templates/conversion_dialog.jinja2
```diff
@@ -398,10 +398,10 @@
                                                     id="parameter_assignment_engraving_mode_precise_btn"
                                                     data-bind="event: {click: updateEngravingMode}"
                                                     class="btn btn-default active">{{ _('Recommended') }}</button>
-                                            <button type="button" value="fast"
+                                            {#<button type="button" value="fast"
```
Yelko-mrbeam, line 402: if we are commenting it we should add a todo  linking to a ticket where we will solve this

Yelko-mrbeam (changes requested):

Yelko-mrbeam:
@khaledsherkawi also take a look at sonar. It's complaining about 2 code smells.

octoprint_mrbeam/static/js/render_fills.js
```diff
@@ -64,6 +66,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
...
+                // TODO: Check if this is valid for any possible combination otherwise remove
```
Yelko-mrbeam: Please create a ticket and link it to the comment so that we plan on doing this

octoprint_mrbeam/static/js/render_fills.js
khaledsherkawi: Done

octoprint_mrbeam/templates/conversion_dialog.jinja2
khaledsherkawi: Done

octoprint_mrbeam/static/js/render_fills.js
khaledsherkawi: Done

khaledsherkawi:
@Yelko-mrbeam  The link attached to the code smells is returning the following:
"The requested project does not exist.
Either it has never been analyzed successfully or it has been deleted."

I believe the code smells are due to the TODOs added in this PR. They consider these to be done before the branch is merged.

irlaec (approved):

JFSolution:
> @Yelko-mrbeam The link attached to the code smells is returning the following: "The requested project does not exist. Either it has never been analyzed successfully or it has been deleted."
> 
> I believe the code smells are due to the TODOs added in this PR. They consider these to be done before the branch is merged.

You need to login with github to see the project
these are the code smells
image

PR #1543, opened 2022-07-19
SW-1308 Add a prefilter and carbon filter maintenance notifications on startup

khaledsherkawi:
- SW-1308 Add a prefilter maintenance notification on startup

- SW-1414 Show parameters with 10% increments (Rounded down to the nearest 10)

### DIFF
--- a/octoprint_mrbeam/static/js/maintenance.js
+++ b/octoprint_mrbeam/static/js/maintenance.js
@@ -71,27 +71,28 @@ $(function () {
             return Math.floor(self.gantryUsage() / 3600);
         });
 
-        self.optimizeParameterPercentageValues = function (val){
+        self.optimizeParameterPercentageValues = function (val) {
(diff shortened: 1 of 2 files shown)


octoprint_mrbeam/static/js/maintenance.js
```diff
@@ -346,6 +347,17 @@ $(function () {
             });
         };
 
+        self.notifyPrefilterMaintenanceRequired = function () {
+            new PNotify({
+                title: gettext("Filter capacity reached 70%"),
+                text: gettext(
+                "At this Level we highly recommend to have a visual check on the filter, to make sure the sensor is representing the fill level of your filter, as it heavily depends on what material you are processing with your device."
```
JFSolution: We don't have a sensor here, this can be missleading. But if the design team said so we can keep it

JFSolution (approved):

JFSolution (approved):

PR #1545, opened 2022-07-26
SW-1523 add possibility to load local instance of the design store

irlaec:


octoprint_mrbeam/static/js/settings/dev_design_store.js
```diff
@@ -10,24 +10,61 @@ $(function () {
...
+
+        self.isLocalhost = ko.computed(function () {
+            return self.selectedEnv() === self.LOCALHOST;
+        });
+
+        self.design_store_staging_iframe_src = ko.computed(function () {
+            return (
+                "https://" +
+                self.devDsVersion() +
+                "-staging-dot-design-store-269610.appspot.com"
```
Yelko-mrbeam: Might be good to make the last part a constant and reuse it on the next block of code or even use the same method. 
protocol = "https://";
appendix = "-dot-design-store-269610.appspot.com";
return (protocol + self.devDsVersion() + "-" + self.selectedEnv() + appendix)
What do you think?

octoprint_mrbeam/static/js/settings/dev_design_store.js
```diff
@@ -66,15 +103,23 @@ $(function () {
         };
 
         self.changeEnv = function () {
-            if (self.selectedEnv() === 'prod') {
-                self.designStore.DESIGN_STORE_IFRAME_SRC = self.DESIGN_STORE_PRODUCTION_IFRAME_SRC;
-            } else if (self.selectedEnv() === 'staging') {
-                self.designStore.DESIGN_STORE_IFRAME_SRC = self.DESIGN_STORE_STAGING_IFRAME_SRC;
-            } else if (self.selectedEnv() === 'dev') {
-                self.designStore.DESIGN_STORE_IFRAME_SRC = self.DESIGN_STORE_DEVELOPMENT_IFRAME_SRC;
+            if (self.selectedEnv() === self.PROD) {
```
Yelko-mrbeam: There are lots of ifs here I think that is avoidable. We could create a map with the options and the results and just call the map with selectedEnv. What do you think?

Yelko-mrbeam (changes requested):

Yelko-mrbeam (approved):

PR #1549, opened 2022-07-28
check if we need to backpropageate

JFSolution:


JFSolution:
this was just to check if we need backpropagation, but the cahnges are already in beta

PR #1560, opened 2022-08-16
SW-1502 run pre-commit on all files (black excluded)

irlaec:


octoprint_mrbeam/static/css/tab_messages.css
```diff
@@ -1,113 +1,140 @@
...
+  .design-img-container {
+  height: calc(70vh);
+}
+.mrb-message-content
+  .single-design-img-gallery
+  .inner-container
+  .design-img-container.single-img-gallery {
+  margin-bottom: 0;
+  height: calc(70vh - 20px);
+}
```
JFSolution, line 85: for me this looks a bit weird, what do you think?

irlaec: As we discussed, Prettier is an opinionated formatter so in general there are few things you can change: https://prettier.io/docs/en/option-philosophy.html
(Same will happen with Black)

JFSolution (approved):

PR #1565, opened 2022-08-19
SW-1711 create setup script for MrBeamPlugin development

irlaec:


Yelko-mrbeam (approved):

ahmed-mrbeam:
I had multiple errors while I tried to run this script, python error, pip not found error, nvm error.
I believe the script needs to be more stable as the Environment setup now will not fit all.

@irlaec  I would suggest to close this one now and  work on the script in a separate issue

PR #1572, opened 2022-09-01
SW-1783 unicode translations fix

Yelko-mrbeam:
Fix for unicode to string casting

### DIFF
--- a/octoprint_mrbeam/services/document_service.py
+++ b/octoprint_mrbeam/services/document_service.py
@@ -47,7 +47,10 @@ class DocumentService:
     def _get_title_translated(self, definition):
         title_key = string_util.separate_camelcase_words(definition.mrbeamdoc_type.value)
         title_translated = gettext(title_key)
-        self._log_error_on_missing_translation(title_key, str(title_translated))
+        if title_translated:
+            self._log_error_on_missing_translation(title_key, title_translated.encode('utf-8', 'ignore'))
+        else:
+            self._logger.error('title_translated is None. This should never be the case. Please check me.')
         return title_translated
 
     @staticmethod
--- a/tests/services/test_burger_menu_service.py
+++ b/tests/services/test_burger_menu_service.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from unittest import TestCase
 
 from mock import patch, MagicMock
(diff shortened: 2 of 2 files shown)


octoprint_mrbeam/services/document_service.py
```diff
@@ -47,7 +47,8 @@ def get_document_simple_for(self, definition, language):
     def _get_title_translated(self, definition):
         title_key = string_util.separate_camelcase_words(definition.mrbeamdoc_type.value)
         title_translated = gettext(title_key)
-        self._log_error_on_missing_translation(title_key, str(title_translated))
+        if title_translated:
```
irlaec, line 50: Would it make sense to log a warning if there's no translation? Or is this a common/valid case?

irlaec (approved):

octoprint_mrbeam/services/document_service.py
Yelko-mrbeam: This is not really a valid case title_translated will be title_key if there is no translation for it. But you are right just in case title_translated is some how None we should log it 👍

JFSolution:
so the error only appeared because of the logging?

tests/services/test_burger_menu_service.py
```diff
@@ -52,3 +53,36 @@ def test_get_burger_menu_model__with_supported_language__should_return_documents
         self.assertIsNot(len(burger_menu_model.documents), 0)
         for document in burger_menu_model.documents:
             self.assertEquals(document.document_link.language, SupportedLanguage.GERMAN)
+
```
JFSolution, line 56: you now only test for spanish, should we test all languages?

Yelko-mrbeam: I think it's fine... the intention of the test is not to test Spanish but something that would have broken before (unicode)... I would include more weird characters if we see it makes sense

Yelko-mrbeam: @Josef-MrBeam I added hopefully all the weird symbols from the 4 languages 😄

Yelko-mrbeam:
> 

Yes, the translated text is unicode and we are trying to verify if this changed with respect of title_key (string)... so we need both as string or both as unicode... And I thought string is better in the long run, in case we change from using english as a key to for example just a normal key string...

JFSolution (approved):

PR #1573, opened 2022-09-05
SW-274 Github actions: setup and test

irlaec:


.github/workflows/main.yml
```diff
@@ -0,0 +1,18 @@
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - alpha
+      - beta
+      - stable
+      - mrbeam2-beta
+      - mrbeam2-stable
```
JFSolution: we don't need these anymore, when this change comes to stable we are not using mrbeam2-* branches anymore

irlaec: Removed

JFSolution (approved):

PR #1583, opened 2022-10-21
SW-1965 Cleanup sonarcloud config files

irlaec:


JFSolution (approved):
was this needed?

irlaec:
MrBeamPlugin also uses the same reusable workflow, so it needs to pass the new arguments as well

PR #1585, opened 2022-11-07
SW-1345 Collect analytics of newly added features

irlaec:


octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, plugin):
         # Job-specific data
         self._current_job_id = None
         self._current_job_time_estimation = -1
```
khaledsherkawi: Please rename all the `self._current_job_time_estimation` into `self._current_job_time_estimation_v1`

khaledsherkawi (changes requested):

octoprint_mrbeam/analytics/analytics_handler.py
khaledsherkawi: and please add a comment to explain which is which (frontend vs backend maybe)

khaledsherkawi (approved):

PR #1586, opened 2022-11-08
SW-2045 Create workflow to run cypress tests and report results to Xray

irlaec:


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,156 @@
...
+    branches:
+      - alpha
+      - beta
+      - stable
+
+env:
+  CYPRESS_TEST_USER_EMAIL: "dev@mr-beam.org"
+  CYPRESS_TEST_USER_PASSWORD: "a"
+  MRBEAM_PLUGIN_URL: "http://plugin-address.org:5002/"
+  MRBEAM_PLUGIN_DOMAIN_NAME: "plugin-address.org"
```
Yelko-mrbeam, line 13: Nice domain name 😌

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,156 @@
...
+          path: /tmp/image.tar
+
+  ui-chrome-tests:
+    runs-on: ubuntu-latest
+    permissions: write-all
+    container:
+      image: cypress/browsers:node16.17.0-chrome106
+      options: --add-host=${{ needs.mrb-docker.outputs.domain }}:host-gateway
+      volumes:
+        - /var/run/docker.sock:/var/run/docker.sock
```
Yelko-mrbeam, line 49: we could add a comment here on why has this been added

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,156 @@
...
+          path: /tmp
+      - name: Load Docker image
+        run: |
+          apt-get --allow-releaseinfo-change update
+          apt-get install docker.io -y
+          docker load --input /tmp/image.tar
+          docker run --name mrbeam-plugin -d -p5002:5000 mrbeam_plugin
+      - run: |
+          envsubst < ./cypress/fixtures/test-data.template.json > ./cypress/fixtures/test-data.json
+          envsubst < ./cypress.config-template.js > ./cypress.config.js
```
Yelko-mrbeam, line 76: should we also include a default cypres.config.js for local usage? I guess that's already part of the repository? 🤔  same for test-data

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,156 @@
...
+      - name: Merge reports
+        run: |
+          npm install --only=dev
+          npm run merge-reports
+      - name: Import to Xray
+        id: xray
+        run: |
+          token=$(curl -H "Content-Type: application/json" -X POST --data '{ "client_id": "${{ secrets.JIRA_CLIENT_ID }}","client_secret": "${{ secrets.JIRA_CLIENT_SECRET }}" }'  https://xray.cloud.getxray.app/api/v2/authenticate| tr -d '"')
+          response=$(curl -H "Content-Type: text/xml" -X POST -H "Authorization: Bearer $token"  --data @"cypress/results/combined.xml" https://xray.cloud.getxray.app/api/v2/import/execution/junit?projectKey=SW)
+          echo "response=$response" >> $GITHUB_OUTPUT
```
Yelko-mrbeam: Nice oauth in 2 lines 😄 Instead of response could we name it with what we really get? I assume something like reports_response or xray_reports_response? not sure what it expects as response if I am honest.. 😓

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,156 @@
...
+      - name: Change issue summary and add link
+        # issue_url: extracts the newly created test execution issue url from the previous response
+        # issue_name: contains the latest commit message. Should be something like "SW-1234 summary"
+        # tests_issue: extracts the "SW-1234" from the commit message, to add it as a link to the text execution issue
```
Yelko-mrbeam, line 128: What happens if the last commit message doesn't have the jira issue? and also even better... what should it happen? Maybe fail the build?

cypress.config-template.js
```diff
@@ -0,0 +1,24 @@
+const { defineConfig } = require("cypress");
+
+module.exports = defineConfig({
+    projectId: "zpa4f8",
```
Yelko-mrbeam, line 4: spa for fate, like 👍

cypress/fixtures/test-data.template.json
```diff
@@ -0,0 +1,7 @@
+{
+  "email": "${CYPRESS_TEST_USER_EMAIL}",
+  "verify_code": "7fd76efb-9ac4-4627-b0b3-3de6a01d07fb",
```
Yelko-mrbeam: is this the designstore's token? might be good to have it as an environment variable as well 🤔

cypress/scripts/cypress-spec-split.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    if (!specFilePaths.length) {
+      throw Error('No spec files found.');
+    }
+
+    const { totalRunners, thisRunner } = getArgs();
+
+    const specsToRun = splitSpecs(specFilePaths, totalRunners, thisRunner);
+
+    console.log(specsToRun.join(','));
```
Yelko-mrbeam, line 121: Nice and easy to understand 👏

Yelko-mrbeam (commented):
Also might be a good idea to squash all the commits into one and remove the comments

.github/workflows/cypress.yml
Yelko-mrbeam: Now that I think about it might be good to add a comment explaining this is just internal 🤔

irlaec: Done

.github/workflows/cypress.yml
irlaec: Done

.github/workflows/cypress.yml
irlaec: Done

.github/workflows/cypress.yml
irlaec: It will fail, but because linking the issue to the test plan will fail. Then the build will fail, but the previous step would have succeeded (which is importing the results to Xray)

.github/workflows/cypress.yml
irlaec: Yes, we have both :smiley:

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,158 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - alpha
+      - beta
+      - stable
+
+env:
+  CYPRESS_TEST_USER_EMAIL: "dev@mr-beam.org"
```
irlaec: Move to secrets

cypress/fixtures/test-data.template.json
irlaec: Done

Yelko-mrbeam (approved):

PR #1594, opened 2022-11-21
SW-1063 SVG generated by the same tool via the same export method are scaled differently

khaledsherkawi:
- Moved "feel lucky" illustrator test to the end and extended it.


khaledsherkawi:
This is a duplicate for #1462

PR #1605, opened 2022-11-28
SW-2296 Pre-commit - format old code

irlaec:


khaledsherkawi:
@irlaec but this is not old code. This was newly introduced. Why wasn't it formatted?

khaledsherkawi (approved):

PR #1607, opened 2022-11-29
SW-664 force to take pictures in camera settings

JFSolution:
- force to take a picture in camera settings
- fix mrb_state live update

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -2324,7 +2324,7 @@ class MrBeamPlugin(
             self.lid_handler._photo_creator.is_initial_calibration = (
                 is_initial_calibration
             )
-            self.lid_handler._startStopCamera("initial_calibration")
+            self.lid_handler._startStopCamera(MrBeamEvents.INITIAL_CALIBRATION)
         succ = self.lid_handler.takeNewPic()
         if succ:
             resp_text = {
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -42,7 +42,7 @@ from ._version import get_versions
 
 __version__ = get_versions()["version"]
 if isinstance(__version__, unicode):
-    __version__ = unicodedata.normalize("NFKD", __version__).encode("ascii", "ignore")
+    __version__ = unicodedata.normalize('NFKD', __version__).encode('ascii', 'ignore')
 
 del get_versions
 
@@ -161,6 +161,7 @@ class MrBeamPlugin(
     TIME_NTP_SYNC_CHECK_FAST_COUNT = 20
     TIME_NTP_SYNC_CHECK_INTERVAL_FAST = 10.0
     TIME_NTP_SYNC_CHECK_INTERVAL_SLOW = 120.0
+
 
     def __init__(self):
         self.mrbeam_plugin_initialized = False
@@ -861,12 +862,10 @@ class MrBeamPlugin(
                 terminalEnabled=self._settings.get(["terminal"]) or self.support_mode,
                 lasersafety_confirmation_dialog_version=self.LASERSAFETY_CONFIRMATION_DIALOG_VERSION,
                 lasersafety_confirmation_dialog_language=language,
-                settings_model=SettingsService(
-                    self._logger, DocumentService(self._logger)
-                ).get_template_settings_model(self.get_model_id()),
-                burger_menu_model=BurgerMenuService(
-                    self._logger, DocumentService(self._logger)
-                ).get_burger_menu_model(self.get_model_id()),
+                settings_model=SettingsService(self._logger, DocumentService(self._logger)).get_template_settings_model(
+                    self.get_model_id()),
+                burger_menu_model=BurgerMenuService(self._logger, DocumentService(self._logger)).get_burger_menu_model(
+                    self.get_model_id()),
                 isDevelop=self.is_dev_env(),
             )
         )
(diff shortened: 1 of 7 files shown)


octoprint_mrbeam/__init__.py
```diff
@@ -2318,8 +2318,11 @@ def take_undistorted_picture(self, is_initial_calibration):
             "New undistorted image is requested. is_initial_calibration: %s",
             is_initial_calibration,
         )
-        self.lid_handler._photo_creator.is_initial_calibration = is_initial_calibration
-        self.lid_handler._startStopCamera("initial_calibration")
+        if is_initial_calibration:
+            self.lid_handler._photo_creator.is_initial_calibration = (
+                is_initial_calibration
+            )
+            self.lid_handler._startStopCamera("initial_calibration")
```
Yelko-mrbeam: Can we create an enum/variable with the event/string? then we can use it on the lid_handler as well as here. Otherwise you really need to know what the method is doing and it could lead to problems soon if we forget to change the string in one of the places...

octoprint_mrbeam/static/js/ready_to_laser_viewmodel.js
```diff
@@ -196,6 +196,15 @@ $(function () {
                 self.fromCurrentData = function (data) {
                     self._fromData(data);
                 };
+
+                self.onDataUpdaterPluginMessage = function (plugin, data) {
+                    if (plugin != "mrbeam") {
```
Yelko-mrbeam: Could we make it a constant somewhere? I think we potentially use it in many places

octoprint_mrbeam/static/js/ready_to_laser_viewmodel.js
```diff
@@ -196,6 +196,15 @@ $(function () {
...
+                        return;
+                    }
+                    if ("mrb_state" in data) {
```
Yelko-mrbeam: same here could we make it a constant? name could be state_key 🤷

octoprint_mrbeam/static/js/settings/camera_settings.js
```diff
@@ -318,6 +320,23 @@ $(function () {
...
+                    self.forceTakeNewPicture()
+                }, 3000);
+            }else{
+                if (self.imageReloadTimer){clearInterval(self.imageReloadTimer)}
+            }
+        }
+
+        self.forceTakeNewPicture = function(){
+            if (self.readyToLaser.lid_fully_open() && !self.readyToLaser.state.interlocksClosed()){
+                OctoPrint.simpleApiCommand("mrbeam", "take_undistorted_picture", {})
```
Yelko-mrbeam: Same here could we make a class or file with all this potential events? it would also help us visualise all the things we use from octoprint. We could start by creating the class and adding this one event on this PR and slowly add more or create a ticket to research other potential events

octoprint_mrbeam/__init__.py
```diff
@@ -860,10 +861,12 @@ def on_ui_render(self, now, request, render_kwargs):
...
-                settings_model=SettingsService(self._logger, DocumentService(self._logger)).get_template_settings_model(
-                    self.get_model_id()),
-                burger_menu_model=BurgerMenuService(self._logger, DocumentService(self._logger)).get_burger_menu_model(
-                    self.get_model_id()),
+                settings_model=SettingsService(
+                    self._logger, DocumentService(self._logger)
+                ).get_template_settings_model(self.get_model_id()),
+                burger_menu_model=BurgerMenuService(
+                    self._logger, DocumentService(self._logger)
+                ).get_burger_menu_model(self.get_model_id()),
```
Yelko-mrbeam: This will give conflicts on my PR 😢

octoprint_mrbeam/static/js/app/view-models/mrbeam-constants.js
```diff
@@ -0,0 +1,4 @@
+const MRBEAM = {
```
Yelko-mrbeam, line 1: 👏

octoprint_mrbeam/static/js/app/view-models/mrbeam-simple-api-commands.js
```diff
@@ -0,0 +1,5 @@
+const SimpleApiCommands = {
```
Yelko-mrbeam, line 1: 👏

octoprint_mrbeam/static/js/ready_to_laser_viewmodel.js
```diff
@@ -196,6 +196,15 @@ $(function () {
                 self.fromCurrentData = function (data) {
                     self._fromData(data);
                 };
+
+                self.onDataUpdaterPluginMessage = function (plugin, data) {
+                    if (plugin !== MRBEAM.PLUGIN_IDENTIFIER) {
+                        return;
+                    }
+                    if (MRBEAM.STATE_KEY in data) {
+                        self._fromData(data, "onDataUpdaterPluginMessage");
```
Yelko-mrbeam, line 205: Shouldn't this one also be part of the events? 🤔 I am not sure why is this needed since in the method doesn't seem to be used or? 😕

octoprint_mrbeam/__init__.py
JFSolution: Ohh I can revert, this was not intended to end up here 😅

octoprint_mrbeam/static/js/ready_to_laser_viewmodel.js
JFSolution: I added "onDataUpdaterPluginMessage" so we know where we got the state from, I would say we can leave this as a string

Yelko-mrbeam (approved):

PR #1609, opened 2022-11-30
SW-2259 fix bug in messaging system

JFSolution:
- fix message shows on first run even if it is disabled

### DIFF
--- a/octoprint_mrbeam/static/js/messages.js
+++ b/octoprint_mrbeam/static/js/messages.js
@@ -230,7 +230,7 @@ $(function () {
                 }
             }
             // not_first_run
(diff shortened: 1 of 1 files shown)


octoprint_mrbeam/static/js/messages.js
```diff
@@ -230,7 +230,7 @@ $(function () {
                 }
             }
             // not_first_run
```
khaledsherkawi, line 232:
```suggestion
            // not_first_run
           // Setting this to false means the message won't show on the first run
```

khaledsherkawi (changes requested):

khaledsherkawi (approved):

PR #1625, opened 2022-12-08
SW-1931 lazy load design store

khaledsherkawi:
SW-1931 Add healthcheck_url to the material store model
 
SW-1931 Add healthcheck_url to the settings service

SW-1931 Remove info logging
 
SW-1931 Add and modify unit tests
 
SW-1931 Modify material store HEAD call to use the healthcheck API
 
SW-1931 Fix a postMessage console error
 
SW-1931 Load design store once and not on every title item click

SW-1931 Initialize variable at the beginning of the method

SW-1931 Modify design store template

SW-1931 Add design store loading spinner styling

SW-1931 Fix bug

SW-1931 Move the design store template into the tabs folder

SW-1931 Lazy load the design store

SW-1931 Modify design store initialization logic
This change will show the network issues page if the device is offline. However, this will not change if the device gets online until the user refreshes the page.

### DIFF
--- a/octoprint_mrbeam/static/js/design_store.js
+++ b/octoprint_mrbeam/static/js/design_store.js
@@ -9,6 +9,7 @@ $(function () {
         // staging:
         // 'https://1-0-0-staging-dot-design-store-269610.appspot.com'
         // 'http://localhost:8080';
+        self.DESIGN_STORE_IFRAME_HEALTHCHECK_SRC = self.DESIGN_STORE_IFRAME_SRC + "/api/healthcheck";
 
         self.loginState = params[0];
         self.navigation = params[1];
@@ -18,8 +19,9 @@ $(function () {
         self.lastUploadedDate = ko.observable("");
         self.eventListenerAdded = ko.observable(false);
 
-        self.onUserLoggedIn = function () {
+        self.initialiseStore = function () {
             self.prepareDesignStoreTab();
+            self.goToStore();
         };
 
         self.getUserSettings = function () {
--- a/octoprint_mrbeam/static/js/design_store.js
+++ b/octoprint_mrbeam/static/js/design_store.js
@@ -20,8 +20,20 @@ $(function () {
         self.eventListenerAdded = ko.observable(false);
 
         self.initialiseStore = function () {
-            self.prepareDesignStoreTab();
-            self.goToStore();
+            let designStoreIframeElement = $("#design_store_iframe");
+            if(designStoreIframeElement.attr("src") !== self.DESIGN_STORE_IFRAME_SRC){
+                self.prepareDesignStoreTab();
+                self.goToStore();
+                // Handle design store if offline
+                // This will show the network issue page if the device is offline
+                // However, if the device gets online afterwards, this will not change
+                // until the user refreshes the page
+                if(!window.mrbeam.isOnline) {
+                    $("#designstore > .loading_spinner_wrapper").hide();
+                    $("#design_store_iframe").hide();
+                    $("#design_store_offline_placeholder").show();
+                }
+            }
         };
 
         self.getUserSettings = function () {
@@ -126,7 +138,7 @@ $(function () {
 
         self.onDiscoveryReceived = function () {
             $("#design_store_iframe").show();
-            $("#design_store_offline_placeholder").hide();
+            $("#designstore > .loading_spinner_wrapper").hide();
 
             // TODO: remove the following Version sanitization once the version
             //  comparative methods support "pep440" versioning (SW-1047)
@@ -284,8 +296,8 @@ $(function () {
             setTimeout(function () {
                 refreshButtonElement.text(refreshButtonText);
             }, 3000);
-            document.getElementById("design_store_iframe").src =
-                self.DESIGN_STORE_IFRAME_SRC;
+            document.getElementById("design_store_iframe").src = "#";
+            self.initialiseStore();
         };
     }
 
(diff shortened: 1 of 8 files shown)


octoprint_mrbeam/static/js/material-store.js
```diff
@@ -70,12 +71,15 @@ $(function () {
                 payload: payload,
             };
 
-            document
-                .getElementById("material_store_iframe")
-                .contentWindow.postMessage(
-                    data,
-                    self.material_store_iframe_src
-                );
+            let materialStoreIframeElement = $("#material_store_iframe");
```
Yelko-mrbeam: if we want to have a variable we could then also initialise it at the beginning of the method. I usually don't like to initialise jquery elements because sometimes it leads to unexpected outcomes.. For example on success of an ajax call sometimes the variable could just be empty for reasons.

octoprint_mrbeam/templates/mrbeam_ui_index.jinja2
```diff
@@ -51,9 +51,9 @@
 					<li class="active"><a href="#workingarea" data-toggle="tab" id="wa_tab_btn">{{ _('working area') }}</a></li>
                     {# Translatots: Name of the tab #}
 					<li><a href="#designlib" data-toggle="tab" id="designlib_tab_btn">{{ _('design library') }}</a></li>
-                    <li><a href="#designstore" data-toggle="tab" id="designstore_tab_btn" onclick="window.mrbeam.viewModels.designStoreViewModel.goToStore()">{{ _('design store') }}</a></li>
+                    <li><a href="#designstore" data-toggle="tab" id="designstore_tab_btn" onclick="window.mrbeam.viewModels.materialStoreViewModel.initialiseStore()">{{ _('design store') }}</a></li>
```
Yelko-mrbeam: This seems wrong.. we are calling initilaliseStore() from materialStoreViewModel... We are also not adding the lazy loading tag on the iframe.... Is this code review ready?

Yelko-mrbeam (changes requested):
Material store seems almost finished. Only one extra detail to be added is the healthcheck_url on the repr() method. Design store code seems to be missing @khaledsherkawi ?

octoprint_mrbeam/static/js/material-store.js
khaledsherkawi: done

octoprint_mrbeam/static/js/design_store.js
```diff
@@ -279,8 +296,8 @@ $(function () {
             setTimeout(function () {
                 refreshButtonElement.text(refreshButtonText);
             }, 3000);
-            document.getElementById("design_store_iframe").src =
-                self.DESIGN_STORE_IFRAME_SRC;
+            document.getElementById("design_store_iframe").src = "#";
```
Yelko-mrbeam, line 299: why are we setting the src to # all the time on reload? 😕 is this working?

octoprint_mrbeam/static/js/design_store.js
```diff
@@ -18,8 +19,21 @@ $(function () {
...
+        self.initialiseStore = function () {
+            let designStoreIframeElement = $("#design_store_iframe");
+            if(designStoreIframeElement.attr("src") !== self.DESIGN_STORE_IFRAME_SRC){
+                self.prepareDesignStoreTab();
+                self.goToStore();
+                // Handle design store if offline
+                // This will show the network issue page if the device is offline
+                // However, if the device gets online afterwards, this will not change
+                // until the user refreshes the page
+                if(!window.mrbeam.isOnline) {
```
Yelko-mrbeam, line 31: 👍

octoprint_mrbeam/static/js/material-store.js
```diff
@@ -65,17 +66,19 @@ $(function () {
...
                 payload: payload,
             };
 
-            document
-                .getElementById("material_store_iframe")
-                .contentWindow.postMessage(
-                    data,
-                    self.material_store_iframe_src
-                );
+            if(materialStoreIframeElement.is(":visible")){
```
Yelko-mrbeam: but if you are calling this method the iframe should be visible right? otherwise you shouldn't call this method... I think it's ok to break here if it's not visible, so that we catch the bug as soon as possible... at least an else  bracket console logging the error

octoprint_mrbeam/static/js/design_store.js
khaledsherkawi: Yes because the page is set not to be reloaded every time we click on the "design store" item in the menu based on the element source. So if we didn't set it to "#" it will not initialise

octoprint_mrbeam/templates/mrbeam_ui_index.jinja2
khaledsherkawi: no it wasn't ready

khaledsherkawi:
> Material store seems almost finished. Only one extra detail to be added is the healthcheck_url on the repr() method. Design store code seems to be missing @khaledsherkawi ?

@Yelko-mrbeam I'm not sure what you mean by that. Is this comment still viable??

octoprint_mrbeam/static/js/material-store.js
khaledsherkawi: resolved

octoprint_mrbeam/static/js/design_store.js
khaledsherkawi: @Yelko-mrbeam

Yelko-mrbeam (approved):

PR #1638, opened 2022-12-20
SW-2442 Enable Github and Jira deployments

ahmed-mrbeam:
* Added deployment-development.yml
* Added deployment-releases.yml

### DIFF
--- a/.github/workflows/deployment-development.yml
+++ b/.github/workflows/deployment-development.yml
@@ -21,7 +21,7 @@ jobs:
       deployments: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - uses: chrnorm/deployment-action@v2
         name: Create GitHub deployment
         id: deployment-dev
--- a/.github/workflows/deployment-development.yml
+++ b/.github/workflows/deployment-development.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches: ["alpha"]
 
-  # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
@@ -17,13 +16,10 @@ permissions:
 jobs:
   deploy-dev:
     name: Deploy to Dev environment
-
     # IMPORTANT: the workflow must have write access to deployments, otherwise the action will fail.
     permissions:
       deployments: write
-
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v1
       - uses: chrnorm/deployment-action@v2
@@ -32,10 +28,6 @@ jobs:
         with:
           token: '${{ github.token }}'
           environment: development
-      # This step could be used in the future to notify or trigger another action
-      - name: Deploy my app
-        run: |
-          # add your deployment code here
 
       - name: Update deployment status (success)
         if: ${{ success() }}
@@ -51,4 +43,12 @@ jobs:
         with:
           token: '${{ github.token }}'
           state: 'failure'
+          deployment-id: ${{ steps.deployment-dev.outputs.deployment_id }}
+
+      - name: Update deployment status (cancelled)
+        if: ${{ cancelled() }}
+        uses: chrnorm/deployment-status@v2
+        with:
+          token: '${{ github.token }}'
+          state: 'cancelled'
           deployment-id: ${{ steps.deployment-dev.outputs.deployment_id }}
--- a/.github/workflows/deployment-development.yml
+++ b/.github/workflows/deployment-development.yml
@@ -1,0 +1,54 @@
+# Deployment workflow to be used for Jira reporting
+name: Deploy alpha to Development
+
+on:
+  push:
+    branches: ["alpha"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+jobs:
+  deploy-dev:
+    name: Deploy to Dev environment
+
+    # IMPORTANT: the workflow must have write access to deployments, otherwise the action will fail.
+    permissions:
+      deployments: write
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - uses: chrnorm/deployment-action@v2
+        name: Create GitHub deployment
+        id: deployment-dev
+        with:
+          token: '${{ github.token }}'
+          environment: development
+
+      - name: Deploy my app
+        run: |
+          # add your deployment code here
+
+      - name: Update deployment status (success)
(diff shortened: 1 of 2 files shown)


.github/workflows/deployment-development.yml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - uses: chrnorm/deployment-action@v2
+        name: Create GitHub deployment
+        id: deployment-dev
+        with:
+          token: '${{ github.token }}'
+          environment: development
+      # This step could be used in the future to notify or trigger another action
+      - name: Deploy my app
```
irlaec: I would suggest to create a ticket to do this, and add a comment with the ticket number here

.github/workflows/deployment-development.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Update deployment status (success)
+        if: ${{ success() }}
+        uses: chrnorm/deployment-status@v2
+        with:
+          token: '${{ github.token }}'
+          state: 'success'
+          deployment-id: ${{ steps.deployment-dev.outputs.deployment_id }}
+
+      - name: Update deployment status (failure)
```
irlaec, line 40: Is this supposed to run if "Update deployment status (success)" fails?

.github/workflows/deployment-releases.yml
```diff
@@ -0,0 +1,89 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - uses: actions-ecosystem/action-regex-match@v2
+        id: regex-match-alpha-release
+        with:
+          text: ${{ github.event.release.name }}
```
irlaec: Does this work if the workflow is manually triggered? Because then the event won't be release but workflow_dispatch, right?

.github/workflows/deployment-development.yml
irlaec: Maybe we could just merge both steps into one, and change the "state" value with an if condition

.github/workflows/deployment-releases.yml
```diff
@@ -0,0 +1,89 @@
...
+          fi
+
+      - uses: chrnorm/deployment-action@v2
+        if: ${{ success() }}
+        name: Create GitHub deployment for the release
+        id: deployment
+        with:
+          token: '${{ github.token }}'
+          environment: ${{ steps.set_release_env.outputs.deployment-env }}
+      # This step could be used in the future to publish release notes or to send Slack notification
```
irlaec: I would suggest the same here as well, to create a ticket and add the id in the comment

.github/workflows/deployment-releases.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Deploy my app
+        run: |
+          echo "Release: ${{ github.event.release.name }}"
+          echo "Release environment : ${{ steps.set_release_env.outputs.deployment-env }}"
+
+      - name: Update deployment status (success)
```
irlaec, line 69: Maybe we could merge these two steps as well?

irlaec (changes requested):

.github/workflows/deployment-development.yml
ahmed-mrbeam: The thing is, I don't have a clear idea yet of what could be done and has a value

.github/workflows/deployment-development.yml
ahmed-mrbeam: The merge would over complicate the checking(discovered that after doing some research) would leave it simple like this and as recommended as well

.github/workflows/deployment-releases.yml
ahmed-mrbeam: Check my answer on the last comment

ahmed-mrbeam: addressed

ahmed-mrbeam:
@irlaec 
Added handling for cancelled state as this one is the only one we would need for now for this workflow

.github/workflows/deployment-development.yml
```diff
@@ -0,0 +1,54 @@
+# Deployment workflow to be used for Jira reporting
+name: Deploy alpha to Development
+
+on:
+  push:
+    branches: ["alpha"]
+
+  workflow_dispatch:
```
irlaec, line 8: Please delete this, as we don't want to run it manually

.github/workflows/deployment-development.yml
irlaec: Ok

.github/workflows/deployment-releases.yml
```diff
@@ -0,0 +1,98 @@
+# Deployment workflow to be used for Jira reporting
+name: Deploy release
+
+on:
+  release:
+    types: [ published ]
+  workflow_dispatch:
```
irlaec, line 7: Also here, if we don't need it

.github/workflows/deployment-development.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  deploy-dev:
+    name: Deploy to Dev environment
+    # IMPORTANT: the workflow must have write access to deployments, otherwise the action will fail.
+    permissions:
+      deployments: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
irlaec: Please change this to use the latest version
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/deployment-releases.yml
```diff
@@ -0,0 +1,98 @@
...
+    name: Deploy a release
+
+    # IMPORTANT: the workflow must have write access to deployments, otherwise the action will fail.
+    permissions:
+      deployments: write
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
```
irlaec: Please update it here as well 
```suggestion
      - uses: actions/checkout@v3
```

irlaec (dismissed):

.github/workflows/deployment-releases.yml
ahmed-mrbeam: changed that one  to use github ref_name

.github/workflows/deployment-development.yml
ahmed-mrbeam: I noticed that some times, the workflow may fail due to github connection error or else. That is why I added this one

.github/workflows/deployment-releases.yml
ahmed-mrbeam: same as the other workflow

.github/workflows/deployment-development.yml
irlaec: Ah, so now we would be able to manually run it as well?

irlaec (approved):

.github/workflows/deployment-development.yml
ahmed-mrbeam: yes, As I'm using github.ref_name to get the tag name and this variable is accessible for all trigger events

PR #1642, opened 2022-12-21
SW-1991 implement new laser head icon

JFSolution:


octoprint_mrbeam/static/js/app/view-models/convert.js
```diff
@@ -585,6 +585,23 @@ $(function () {
                 Object.keys(tmp).length
             );
             self.custom_materials(tmp);
+            $(document).on("mouseover", ".material_entry", function () {
+                var $this = $(this);
```
irlaec: I would suggest to rename `$this ` with a significant name, like for example if it is the material entry or whatever, or to otherwise use `$(this)` in the rest of the code

irlaec (approved):

octoprint_mrbeam/static/js/app/view-models/convert.js
```diff
@@ -585,6 +585,23 @@ $(function () {
...
+                        material_name_DOMelement.scrollWidth &&
+                    !material_name_element.attr("title")
+                ) {
+                    $this.tooltip({
+                        title: material_name_element.text(),
+                        placement: "top",
+                    });
+                    $this.tooltip("show");
+                }
+            });
```
Josef-MrBeam, line 604:
```suggestion
$(document).on("mouseover", ".material_entry", function () {
                var material_entry_element = $(this);
                var material_name_DOMelement =
                    material_entry_element.context.getElementsByClassName("material_name")[0];
                var material_name_element = $(material_name_DOMelement);
                if (
                    material_name_DOMelement.offsetWidth <
                        material_name_DOMelement.scrollWidth &&
                    !material_name_element.attr("title")
                ) {
                    material_entry_element.tooltip({
                        title: material_name_element.text(),
                        placement: "top",
                    });
                    material_entry_element.tooltip("show");
                }
            });
```

octoprint_mrbeam/static/js/app/view-models/convert.js
JFSolution: thx for the feedback, I added a name

PR #1643, opened 2022-12-22
SW-2424-disable-logging-for-pressure-sensor

JFSolution:
disables the debug logging for legacy devices with a migration


tests/migrations/test-migration-Mig004.py
```diff
@@ -0,0 +1,41 @@
+import os
```
irlaec: Please change the module name to comply with PEP8 (underscores instead of dashes)
https://peps.python.org/pep-0008/#package-and-module-names

tests/migrations/test-migration-Mig004.py
```diff
@@ -0,0 +1,41 @@
...
+
+from mock import patch
+
+from octoprint_mrbeam.migration import Mig004DisableDebugLogging
+import unittest
+
+
+class TestMigrationMig004(unittest.TestCase):
```
irlaec: No need to change now, but I thought we agreed to use pytest :thinking:

irlaec (dismissed):

tests/migrations/test-migration-Mig004.py
JFSolution: I copied the existing test

tests/migrations/test-migration-Mig004.py
JFSolution: ok

irlaec (approved):

PR #1645, opened 2022-12-23
SW-2303-show-air-filter-version-and-serial-in-settings

JFSolution:
- order mrb_state dict
- add air filter class
- return air filter serial modelId and model name in debug and about settings page

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -677,6 +677,7 @@ class MrBeamPlugin(
                 "js/app/view-models/settings/camera.js",
                 "js/app/view-models/settings/backlash.js",
                 "js/app/view-models/settings/leds.js",
+                "js/app/view-models/settings/about.js",
                 "js/app/helpers/path-magic.js",
                 "js/lib/simplify.js",
                 "js/lib/clipper.js",
@@ -844,9 +845,6 @@ class MrBeamPlugin(
                 laserhead_model=self.laserhead_handler.get_current_used_lh_data()[
                     "model"
                 ],
-                airfilter_serial=self.airfilter.serial,
-                airfilter_model=self.airfilter.model,
-                airfilter_model_id=self.airfilter.model_id,
                 env=self.get_env(),
                 mac_addrs=self._get_mac_addresses(),
                 env_local=self.get_env(self.ENV_LOCAL),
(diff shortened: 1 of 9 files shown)


octoprint_mrbeam/__init__.py
```diff
@@ -36,6 +36,7 @@
 
 from octoprint_mrbeam.rest_handler.update_handler import UpdateRestHandlerMixin
 from octoprint_mrbeam.util.connectivity_checker import ConnectivityChecker
+from .iobeam.airfilter import airfilter
```
irlaec: Can we change this from relative to absolute import?

octoprint_mrbeam/iobeam/airfilter.py
```diff
@@ -0,0 +1,81 @@
...
+        if self._pressure2 is not None:
+            return {
+                "pressure1": self._pressure1,
+                "pressure2": self._pressure2,
+                "pressure3": self._pressure3,
+                "pressure4": self._pressure4,
+            }
+        return self._pressure1
+
+    def set_serial(self, serial):
```
irlaec: Have you considered using a decorator the same way you do with the getter? Something like
```suggestion
    @serial.setter
    def serial(self, serial):
```

tests/iobeam/test_aifilter.py
```diff
@@ -0,0 +1,94 @@
+from octoprint_mrbeam.iobeam.airfilter import AirFilter
+
+
+def test_model_name():
```
irlaec: I would suggest to split the unit tests that contain more than one assertion in different tests

octoprint_mrbeam/iobeam/airfilter.py
```diff
@@ -0,0 +1,81 @@
+# singleton
+from octoprint_mrbeam.mrb_logger import mrb_logger
```
irlaec, line 2: Please add docstrings to the new methods

irlaec (changes requested):

octoprint_mrbeam/__init__.py
JFSolution: sure

octoprint_mrbeam/iobeam/airfilter.py
JFSolution: I will change to the setter decorator

tests/iobeam/test_aifilter.py
JFSolution: But it is all testing the same functionality, thats why I grouped it

octoprint_mrbeam/iobeam/airfilter.py
JFSolution: Ohh yes I missed that, thanks for the hint.

tests/iobeam/test_aifilter.py
irlaec: But we are testing different scenarios (more than one AAA) so that's why I would separate it, like stated [here](https://www.educative.io/blog/unit-testing-best-practices-overview#:~:text=7.%20One%20use%20case%20per%20unit%20test). But maybe it's just a matter of preference :)

irlaec (approved):

PR #1651, opened 2023-01-17
SW-2300 detect air filter iii in mr beam plugin

JFSolution:
- detect air filter in Mr Beam Plugin
- switch profile of the air filter
- use profiles for maintenance parameters
- use profiles for maintenance page
- fix shopify links


octoprint_mrbeam/profiles/airfilter_system/airfilter_system_id_8.yaml
```diff
@@ -1,6 +1,11 @@
 #airfilter 3 system
+prefilter_stages: 2
+carbonfilter_stages: 1
 prefilter:
-  - lifespan: 40 #TODO
-  - lifespan: 40 #TODO
-carbon_filter:
-  - lifespan: 280 #TODO
+  - lifespan: 80 #TODO
```
irlaec: Could you please add a explanation to these TODO and maybe a ticket number?

octoprint_mrbeam/static/js/app/view-models/settings/maintenance.js
```diff
@@ -356,6 +353,34 @@ $(function () {
...
+        self.shopifyLink = function (stagename, stageid) {
+            let link;
+            if (stagename === "prefilter") {
+                link = self.prefilterShopify()[stageid];
+            } else if (stagename === "carbonfilter") {
+                link = self.carbonfilterShopify()[stageid];
+            } else {
+                link = null;
+            }
+            if (link != null) {
```
irlaec: Wouldn't be enough by just returning link? We already assign it a null value if it doesn't have one.
I think we could just remove this second if-else

tests/iobeam/test_aifilter.py
```diff
@@ -1,11 +1,28 @@
+import os
+
 from mock.mock import MagicMock
 
-from octoprint_mrbeam.iobeam.airfilter import AirFilter
+from octoprint_mrbeam.iobeam.airfilter import AirFilter, airfilter
+
+pluginmock = MagicMock(
```
irlaec: We have a fixture for the MrBeamPlugin (in conftest), I don't know if it would make sense to use that instead

irlaec (changes requested):

octoprint_mrbeam/static/js/app/view-models/settings/maintenance.js
JFSolution: You are right

octoprint_mrbeam/profiles/airfilter_system/airfilter_system_id_8.yaml
JFSolution: sure

tests/iobeam/test_aifilter.py
JFSolution: I changed it to use this instead, thanks for the tip

irlaec (approved):
I just noticed the name of the test file has a typo --> test_aiRfilter.py

PR #1658, opened 2023-02-06
SW-2079 implement new material settings

JFSolution:
changed parser to use dreamcut x values
add first draft of material settings


octoprint_mrbeam/util/material_csv_parser.py
```diff
@@ -4,42 +4,33 @@
 MRBEAM = "Mr Beam II"
 MRB_DREAMCUT = "MrB II Dreamcut"
 MRB_DREAMCUT_S = "MrB II Dreamcut S"
+MRB_DREAMCUT_x = "MrB II Dreamcut x"
```
khaledsherkawi: Why is the "x" not capitalized as in the "S"?

tests/util/test_material_settings_parser.py
```diff
@@ -0,0 +1,71 @@
...
+        .get("Kraftplex")
+        .get("colors")
+        .get("795f39")
+        .get("cut")[2]
+        .get("cut_f")
+        != 350
+    )
+    # diff = material_x.keys() - material_s.keys()
+    print(material_x, material_s)
+    # assert False
```
khaledsherkawi: Are these comments intended to stay here?

octoprint_mrbeam/util/material_csv_parser.py
```diff
@@ -4,42 +4,33 @@
...
 DEFAULT_LASER_MODEL = "0"
 LASER_MODEL_S = "S"
+LASER_MODEL_x = "x"
+
+VALID_VERSIONS = [MRBEAM, MRB_DREAMCUT, MRB_DREAMCUT_S, MRB_READY, MRB_DREAMCUT_x]
 
 
 def model_ids_to_csv_name(device_model_id, laser_model_id):
     convert = {
+        (device_info.MODEL_MRBEAM_2_DC_x, LASER_MODEL_x): MRB_DREAMCUT_x,
```
khaledsherkawi: Maybe we can get rid of this mapping and just use the laser head model ID for the CSV name excluding only the DC_R1 and DC_R2 model handling. But this is optional.

khaledsherkawi (changes requested):

octoprint_mrbeam/util/material_csv_parser.py
JFSolution: because the x should be small from the naming convention

octoprint_mrbeam/util/material_csv_parser.py
JFSolution: I will not change it as this will change with the new material settings anyway

tests/util/test_material_settings_parser.py
JFSolution: will remove

tests/util/test_material_settings_parser.py
```diff
@@ -0,0 +1,67 @@
+from octoprint_mrbeam import parse_csv
+from octoprint_mrbeam.util.device_info import (
+    MODEL_MRBEAM_2_DC,
+    MODEL_MRBEAM_2_DC_x,
```
khaledsherkawi: this still needs to be 'MODEL_MRBEAM_2_DC_X'

khaledsherkawi: I would say even if the naming convention is small x but as a representation it still make sense for this to be capitalized

khaledsherkawi (changes requested):

tests/util/test_material_settings_parser.py
JFSolution: Yes, for the constants it makes sense, I missed this one

khaledsherkawi (approved):

PR #1679, opened 2023-02-16
SW-2720 resolve conflicts for beta release

JFSolution:
resolve the critical and major bugs and the duplicate code

### DIFF
--- a/octoprint_mrbeam/static/css/svgtogcode.css
+++ b/octoprint_mrbeam/static/css/svgtogcode.css
@@ -588,7 +588,6 @@ li.material_entry.custom.incompatible_material::after {
   content: "";
   display: block;
   height: 0.5em;
-  background-color: white;
   position: relative;
   left: -1%;
   top: -0.35em;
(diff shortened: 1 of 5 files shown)


octoprint_mrbeam/static/css/svgtogcode.css
```diff
@@ -588,7 +588,6 @@ li.material_entry.custom.incompatible_material::after {
   content: "";
   display: block;
   height: 0.5em;
-  background-color: white;
   position: relative;
```
khaledsherkawi, line 591: Better change the 'background' CSS into 'background-color' as well

khaledsherkawi (dismissed):

octoprint_mrbeam/static/css/svgtogcode.css
JFSolution: This need to be background as it will not work as background-color

khaledsherkawi (approved):

PR #1684, opened 2023-02-20
SW-2810 Zero width strokes are ignored when added to working area

khaledsherkawi:
SW-2810 Fix bug resulting in ignoring unattached vertical strokes

### DIFF
--- a/octoprint_mrbeam/static/js/app/view-models/working-area.js
+++ b/octoprint_mrbeam/static/js/app/view-models/working-area.js
@@ -344,9 +344,7 @@ $(function () {
                 let col = items[i].attr()[color_attr];
                 const bb = items[i].getBBox();
                 if (
-                    bb.w > 0 &&
-                    // TODO: the line path has zero height so this should be done in a different manner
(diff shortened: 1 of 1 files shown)


octoprint_mrbeam/static/js/app/view-models/working-area.js
```diff
@@ -344,9 +344,7 @@ $(function () {
                 let col = items[i].attr()[color_attr];
                 const bb = items[i].getBBox();
                 if (
-                    bb.w > 0 &&
-                    // TODO: the line path has zero height so this should be done in a different manner
-                    // bb.h > 0 && // filters elements without dimension, e.g. <path> without d attrib.
+                    (bb.w > 0 || bb.h > 0) &&
```
JFSolution, line 347: why was this not done before? If it is already in a comment?

JFSolution (approved):

octoprint_mrbeam/static/js/app/view-models/working-area.js
khaledsherkawi: I'm not really sure what went there. But I think there was an attempt to remove elements with no bb but it was poorly executed

PR #1687, opened 2023-02-26
feature/SW-2537-ds-show-designstore-notification-on-first-laserhead-x-detection

khaledsherkawi:
SW-2588 Remove current handling of the design store 'new designs' notification
SW-2588 Refactor code
SW-2588 Modify notification dot color to match Mr Beam orange color
SW-2588 Add TODOs for related tickets
SW-2537 Refactor laserhead-changed.js
SW-2537 Refactor design-store.js
SW-2537 Send event and show notify icon on laserhead change to X detection
SW-2537 Revamp method


octoprint_mrbeam/static/js/app/view-models/design-store.js
```diff
@@ -11,23 +11,31 @@ $(function () {
         // 'http://localhost:8080';
         self.DESIGN_STORE_IFRAME_HEALTHCHECK_SRC =
             self.DESIGN_STORE_IFRAME_SRC + "/api/healthcheck";
+        self.DESIGN_STORE_TAB_ELEMENT = $("#designstore_tab_btn");
```
Yelko-mrbeam, line 14: I like the idea of storing a reference to the tab button here, although this reference could be lost if for some reason the object is detached of the DOM right? Somehow I've seen this behaviour in our plugin before... I think it's safer to always look for the element like $('#element'). What we could do is save the reference to the button id and reuse it. What do you think @khaledsherkawi ?

octoprint_mrbeam/static/js/app/view-models/design-store.js
```diff
@@ -162,6 +170,13 @@ $(function () {
             };
 
             self.sendMessageToDesignStoreIframe("userData", userData);
+
+            if (self.laserheadChangedVM.isLaserheadChanged()) {
+                self.sendMessageToDesignStoreIframe("laserheadChanged", {
```
Yelko-mrbeam: I think the laserhead change event should affect only the Mr Beam plugin. We could send to the designs store an event to select the filter though. Something like `self.sendMessageToDesignStoreIframe("selectFilter" {recommendedBy:  'laserheadModel', value: 'x'})`

octoprint_mrbeam/static/js/app/view-models/design-store.js
```diff
@@ -162,6 +170,13 @@ $(function () {
             };
 
             self.sendMessageToDesignStoreIframe("userData", userData);
+
+            if (self.laserheadChangedVM.isLaserheadChanged()) {
```
Yelko-mrbeam: I think we only want to send the event if the laserhead x is the one selected so far right?

octoprint_mrbeam/static/js/app/view-models/modal/laserhead-changed.js
```diff
@@ -59,19 +59,31 @@ $(function () {
...
         };
 
+        self.isLaserheadChanged = function () {
+            return (
+                self.loginState.currentUser?.()?.active &&
+                self.settings.settings.plugins.mrbeam.laserheadChanged()
+            );
+        };
+
+        self.laserheadXDetectedForFirstTime = function () {
```
Yelko-mrbeam: This is only checking if the laserhead was changed to a laserhead x right? It doesn't contain information about the first time this happened or am I wrong? I would then propose to rename the function. I would vote to not overcomplicate it and leave it with this logic without taking into account if it's the first time it was changed or not but just if it was changed to laserhead x from any different laserhead

octoprint_mrbeam/static/js/app/view-models/design-store.js
```diff
@@ -11,23 +11,31 @@ $(function () {
...
         self.loginState = params[0];
         self.navigation = params[1];
         self.analytics = params[2];
         self.settings = params[3];
+        self.laserheadChangedVM = params[4];
 
         self.lastUploadedDate = ko.observable("");
         self.eventListenerAdded = ko.observable(false);
 
+        self.onUserLoggedIn = function () {
```
Yelko-mrbeam, line 25: when is this being called?

Yelko-mrbeam (changes requested):

octoprint_mrbeam/static/js/app/view-models/design-store.js
khaledsherkawi: Oh, yea that's 100% the case and that is why I didn't do that in the material store as the material store tab is dynamically rendered. However, the design store tab is not so it is always attached to the DOM. If we change the behavior to match the one of the material store then this needs to be changed as well.

octoprint_mrbeam/static/js/app/view-models/design-store.js
khaledsherkawi: I like that; I will implement it

octoprint_mrbeam/static/js/app/view-models/design-store.js
khaledsherkawi: Yea the idea was to keep the handling of the detected laser head to the design store if we decided to also tag S laser head designs or for any future laser heads; However we can limit it to the X laser head for now.

octoprint_mrbeam/static/js/app/view-models/modal/laserhead-changed.js
khaledsherkawi: I will change to this:
```suggestion
        self.laserheadXDetected = function () {
```

octoprint_mrbeam/static/js/app/view-models/design-store.js
khaledsherkawi: When the user logs in or after page loads and the user is already logged in

octoprint_mrbeam/static/js/app/view-models/design-store.js
Yelko-mrbeam: now that I think about it maybe more generic like:
`"selectFilter",{type:"recommendedBy", subsection:  'laserheadModel', value: 'x'}`

khaledsherkawi: Done

octoprint_mrbeam/static/js/app/view-models/modal/laserhead-changed.js
khaledsherkawi: Done

octoprint_mrbeam/static/js/app/view-models/design-store.js
khaledsherkawi: done

Yelko-mrbeam (approved):

PR #1688, opened 2023-02-26
SW-2236 MS - show Material store notification for laserhead x first detection

khaledsherkawi:
SW-2236 Move notification icon logic to main mr beam view model
SW-2236 Fix spelling mistake
SW-2236 Add material store notification styling
SW-2236 Add material store logic for laserhead model detection


octoprint_mrbeam/templates/mrbeam_ui_index.jinja2
```diff
@@ -53,7 +53,7 @@
 					<li><a href="#designlib" data-toggle="tab" id="designlib_tab_btn" data-test="mrbeam-ui-index-design-library">{{ _('design library') }}</a></li>
                     <li><a href="#designstore" data-toggle="tab" id="designstore_tab_btn" onclick="window.mrbeam.viewModels.designStoreViewModel.initialiseStore()" data-test="mrbeam-ui-index-design-store">{{ _('design store') }}</a></li>
                     {% if settings_model.material_store.enabled %}
-                        <li><a href="#material_store_content" data-toggle="tab" id="materialsotre_tab_btn" onclick="window.mrbeam.viewModels.materialStoreViewModel.initialiseStore('{{ settings_model.material_store.healthcheck_url }}', '{{ settings_model.material_store.url }}')">{{ _('material store') }}</a></li>
+                        <li><a href="#material_store_content" data-toggle="tab" id="materialstore_tab_btn" onclick="window.mrbeam.viewModels.materialStoreViewModel.initialiseStore('{{ settings_model.material_store.healthcheck_url }}', '{{ settings_model.material_store.url }}')">{{ _('material store') }}</a></li>
```
Yelko-mrbeam, line 56: 👍

octoprint_mrbeam/static/js/app/view-models/mrbeam.js
```diff
@@ -629,6 +637,19 @@ $(function () {
                 return c.action === "restart_safe";
             });
         };
+
+        self.showNotifyIcon = function (element) {
```
Yelko-mrbeam, line 641: 👏

octoprint_mrbeam/static/js/app/view-models/material-store.js
```diff
@@ -36,12 +40,21 @@ $(function () {
...
         };
 
         self.sendInitDetailsToMaterialStoreIframe = function () {
             self.sendMessageToMaterialStoreIframe(
                 MATERIAL_STORE_EVENT_TYPE.MR_BEAM_LOAD
             );
+
+            // send new laserhead model ID if changed
+            if (self.laserheadChangedVM.isLaserheadChanged()) {
+                self.sendMessageToMaterialStoreIframe("laserheadChanged", {
```
Yelko-mrbeam: Let's send the message here the same way we do it in the design store. for example: `event = 'selectFilter', payload = 'laserheadX'}`

Yelko-mrbeam (changes requested):

octoprint_mrbeam/static/js/app/view-models/material-store.js
Yelko-mrbeam: although check the material store frontend... I think we might have the event already defined there

khaledsherkawi: I checked but the current event is to only select product and not a filter; This is done.

octoprint_mrbeam/static/js/app/view-models/material-store.js
```diff
@@ -36,12 +41,25 @@ $(function () {
...
             );
+
+            // send new laserhead model ID if changed
+            if (self.laserheadChangedVM.laserheadXDetected()) {
+                self.sendMessageToMaterialStoreIframe(
+                    MATERIAL_STORE_EVENT_TYPE.SELECT_FILTER,
+                    {
+                        type: "recommendedBy",
+                        subsection: "laserheadModel",
+                        value: "x",
```
Yelko-mrbeam: On the material store it will be an actual filter, not recommendation and as far as I know we can only filter by "tags", "categories" or "metadata". I think on the mrbeam plugin we don't need to know that so we could just send the name of the filter like: value: 'laserheadx'. We could keep an enum with the expected available filters for the material store on the Mr Beam plugin, sort of what we have with MATERIAL_STORE_EVENT_TYPE. 
Shall we talk about it?

khaledsherkawi: done

Yelko-mrbeam (approved):

PR #1689, opened 2023-02-28
SW-2539 Implement design/text changes for Page 1 (start)

khaledsherkawi:
SW-2539 Modify first page of the welcome wizard


octoprint_mrbeam/templates/wizard/firstrun_start.jinja2
```diff
@@ -1,9 +1,19 @@
...
+        Please read the following steps to complete the setup!
+    </h4>
+    <br>
+    <p>
+        This dialog will get your Mr Beam ready in just a couple of minutes.<br>
+        Afterwards you can take a tour that will guide you through your first laser job.<br>
+    <br>
+        Please have a read, and if anything is unclear, get in touch!
+    </p>
+    {% endtrans %}
```
JFSolution, line 18: Why we use here a different way to translate 😅

JFSolution (approved):

octoprint_mrbeam/templates/wizard/firstrun_start.jinja2
khaledsherkawi: I would rather this as it is much easier to translate than to look for every single string individually

PR #1694, opened 2023-03-07
SW-2806 laser head lifespan switching with model

JFSolution:
- added lifespan to the laser head profiles
- added unittest
- get lifespan of laser head from backend

### DIFF
--- a/octoprint_mrbeam/iobeam/laserhead_handler.py
+++ b/octoprint_mrbeam/iobeam/laserhead_handler.py
@@ -11,6 +11,7 @@ LASERHEAD_MAX_TEMP_FALLBACK = 55.0
 LASERHEAD_MAX_DUST_FACTOR_FALLBACK = 3.0 # selected the highest factor
 LASERHEAD_MAX_CORRECTION_FACTOR_FALLBACK = 1
 LASERHEAD_MAX_INSTENSITY_INCLUDING_CORRECTION_FALLBACK = 1500
+LASERHEAD_LIFESPAN_FALLBACK = 40
 
 LASERHEAD_STOCK_ID = 0
 LASERHEAD_S_ID = 1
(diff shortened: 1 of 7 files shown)


octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -652,3 +653,36 @@ def default_laserhead_max_intensity_including_correction(self):
...
+        self._logger.debug(
+            "Current Laserhead lifespan:{}".format(current_laserhead_properties["lifespan"]))
+        return current_laserhead_properties["lifespan"]
+
+    @property
+    def default_laserhead_lifespan(self):
+        """Default lifespan for laser head. To be used by other modules at init time.
+
+        Returns:
+            float: Laser head default lifespan
```
irlaec: If we expect the return value to be float, `LASERHEAD_LIFESPAN_FALLBACK` should be 40.0. Otherwise please change the docstring

irlaec (changes requested):

octoprint_mrbeam/iobeam/laserhead_handler.py
JFSolution: Ohh that was a copy past error, will fix it

octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -652,3 +653,36 @@ def default_laserhead_max_intensity_including_correction(self):
         """
 
         return LASERHEAD_MAX_INSTENSITY_INCLUDING_CORRECTION_FALLBACK
+
+    @property
+    def current_laserhead_lifespan(self):
+        """Return the current laser head lifespan.
+
+        Returns:
+            float: Laser head lifespan
```
JFSolution:
```suggestion
            int: Laser head lifespan
```

octoprint_mrbeam/iobeam/laserhead_handler.py
JFSolution:
```suggestion
            int: Laser head default lifespan
```

irlaec (approved):

PR #1697, opened 2023-03-09
SW-2786 filter temperature values

JFSolution:
* added fire warning to analytics
* added temperature filter to filter spikes


JFSolution:
This will be implemented in iobeam instead
https://bitbucket.org/mrbeam/iobeam/pull-requests/83

PR #1701, opened 2023-03-10
SW-2541 Implement design/text changes for Page 4 (laser safety)

khaledsherkawi:
SW-254 Add new laser safety design


octoprint_mrbeam/__init__.py
```diff
@@ -139,7 +139,7 @@ class MrBeamPlugin(
     ENV_LASER_SAFETY = "laser_safety"
     ENV_ANALYTICS = "analytics"
 
-    LASERSAFETY_CONFIRMATION_DIALOG_VERSION = "0.4"
+    LASERSAFETY_CONFIRMATION_DIALOG_VERSION = "0.5"
```
JFSolution, line 142: Will this show the dialoge angain for all that already confirmed it?

JFSolution (approved):

octoprint_mrbeam/__init__.py
khaledsherkawi: This basically is just for documentation and nothing else

khaledsherkawi: It is used in the payload uploaded to the cloud but it won't trigger another confirmation

PR #1703, opened 2023-03-13
SW-2571 implement a setting if the heavy duty pre filter is used

JFSolution:
added cypress test
added unittest
added choice input with a tooltip for switching the prefilter type


cypress/e2e/settings/maintenance.cy.js
```diff
@@ -96,4 +106,50 @@ describe("Maintenance", function () {
             "not.visible"
         );
     });
+
+    it("When checkbox is clicked then setting is saved.", function () {
+        cy.get('[id="settings_plugin_mrbeam_maintenance_link"]').click();
+        // cy.request('/settings').its('body').should('include', '<h1>Admin</h1>')
```
khaledsherkawi:
```suggestion
```

octoprint_mrbeam/static/js/app/view-models/settings/maintenance.js
```diff
@@ -196,6 +223,42 @@ $(function () {
...
+                    if (clickedElement.id === "prefilter_shop_link") {
+                        // Prevent the click event from propagating further
+                        event.stopPropagation();
+                    } else {
+                        self.heavyDutyPrefilter(
+                            radioInput.getAttribute("value")
+                        );
+                    }
+                });
+            });
```
khaledsherkawi, line 268: This is quite unclear; What are we trying to do here?
And maybe we can move this chunk and the one below into different functions for readability purposes so that self.onAllBound is more clear

khaledsherkawi (changes requested):

octoprint_mrbeam/static/js/app/view-models/settings/maintenance.js
JFSolution: the purpose is to make the whole div clickable to change the radio selection instead just the radio input it selfe

khaledsherkawi: Yea maybe just move this chunk and the one below into separate methods with readable names for more clarity

khaledsherkawi (changes requested):

octoprint_mrbeam/static/js/app/view-models/settings/maintenance.js
```diff
@@ -196,6 +223,49 @@ $(function () {
                 });
             });
             self.updateSettingsAbout();
+
+            self._make_prefilter_elements_clickable();
+            self._add_tooltip_for_prefilter_title();
```
khaledsherkawi:
```suggestion
            self._makePrefilterElementsClickable();
            self._addTooltipForPrefilterTitle();
```

octoprint_mrbeam/static/js/app/view-models/settings/maintenance.js
```diff
@@ -196,6 +223,49 @@ $(function () {
...
+        };
+
+        self._add_tooltip_for_prefilter_title = function (element) {
```
khaledsherkawi:
```suggestion
        self. _addTooltipForPrefilterTitle = function (element) {
```

octoprint_mrbeam/static/js/app/view-models/settings/maintenance.js
```diff
@@ -196,6 +223,49 @@ $(function () {
...
+                material_entry_element.tooltip({
+                    title: "<img src='" + image + "' width='300px'>",
+                    placement: "right",
+                    html: true,
+                });
+                material_entry_element.tooltip("show");
+            });
+        };
+
+        self._make_prefilter_elements_clickable = function () {
```
khaledsherkawi:
```suggestion
        self. _makePrefilterElementsClickable = function () {
```

khaledsherkawi (changes requested):

khaledsherkawi (dismissed):

khaledsherkawi (approved):

PR #1722, opened 2023-03-21
SW-1256 laser will think a job is done if you open the lid while moving the laserhead

JFSolution:
- unpause unintended pause
- refactor use of time of payload with fallback value

### DIFF
--- a/octoprint_mrbeam/printing/comm_acc2.py
+++ b/octoprint_mrbeam/printing/comm_acc2.py
@@ -2586,7 +2586,6 @@ class MachineCom(object):
             payload = {}
         else:
             payload = self._get_printing_file_state()
-        # payload = self._get_printing_file_state()
         payload["trigger"] = trigger
         self._logger.debug(
             "pause %s send_cmd: %s trigger: %s force: %s ispaused: %s",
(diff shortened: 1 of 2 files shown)


octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -2564,27 +2570,52 @@ def cancelPrint(self, failed=False, error_msg=False):
...
+                    "cycle_start"
+                ] = True  # sends realtime command to resume print
+            if self.isPaused():
+                # continue from intentional pause
+                if self._pauseWaitStartTime:
+                    self._pauseWaitTimeLost = self._pauseWaitTimeLost + (
+                        time.time() - self._pauseWaitStartTime
+                    )
+                    self._pauseWaitStartTime = None
+                self._pause_delay_time = time.time()  # get_uptime()
```
irlaec: Are we planning to use get_uptime() in the future, or what is this comment about?

irlaec (dismissed):

octoprint_mrbeam/printing/comm_acc2.py
JFSolution: ohh yes, I thought about already using it there but then I need to change it everywhere, I can remove this comment as it is old

JFSolution:
```suggestion
                self._pause_delay_time = time.time()
```

irlaec (approved):

PR #1728, opened 2023-03-24
SW-2950 Check what to replace analytics with in the Mr Beam plugin

khaledsherkawi:
SW-2950 Fix texts


octoprint_mrbeam/templates/settings/analytics_settings.jinja2
```diff
@@ -13,7 +13,7 @@
                         {{ _('We know that your Mr Beam is important to you, and we want to make it even better. That\'s why we\'re asking for your help in improving it by sharing optional diagnostic and usage data. These data would allow us to better understand your needs and preferences, leading to faster identification and resolution of issues, and ultimately, improved features and functionality.') }}
                     <br>
                     <br>
-                        {{ _('We understand that your privacy is crucial, and we want to assure you that none of the data collected includes your personal information such as your name, login details, designs, or any other private content. You are in full control, and you can change your mind at any time by going to <strong>"Settings > Better together"</strong>.') }}
+                        {{ _('We understand that your privacy is crucial, and we want to assure you that none of the data collected includes your personal information such as your name, login details, designs, or any other private content.') }}
```
JFSolution, line 16: If you change this in to separate strings we don't need to translate it twice

JFSolution (changes requested):
split tranlsatet string in two so we don't need to translate it twice

octoprint_mrbeam/templates/settings/analytics_settings.jinja2
khaledsherkawi: Done

JFSolution (approved):

PR #1729, opened 2023-03-27
SW-2923 send pcf anomaly error to backend

JFSolution:
- request hardware malfunctions on print start
- handling of iobeam errors added
- show error when job failed
- light leds in error when job failed
- notifications can be dismissed
- added error codes to notifications

- fix exceptions

### DIFF
--- a/octoprint_mrbeam/iobeam/hw_malfunction_handler.py
+++ b/octoprint_mrbeam/iobeam/hw_malfunction_handler.py
@@ -21,6 +21,32 @@ def hwMalfunctionHandler(plugin):
     return _instance
 
 
+class HwMalfunction:
+    def __init__(self, malfunction_id, msg, data, error_code=None, priority=0):
+        self.id = malfunction_id
+        self._msg = msg
+        self.data = data
+        self.priority = priority
+        self.error_code = error_code
+
+    @property
+    def msg(self):
+        """Returns the message to be shown to the user."""
+        # if error_code is set, only show the code else show the message
+        if self.error_code:
+            return "Error Code: {}".format(self.error_code)
+
+        return self._msg
+
+    def __str__(self):
+        return "HwMalfunction(id:{id}, msg:{msg}, data:{data}, prio:{priority})".format(
+            id=self.id, msg=self.msg, data=self.data, priority=self.priority
+        )
+
+    def __repr__(self):
+        return self.__str__()
+
+
 class HwMalfunctionHandler(object):
     MALFUNCTION_ID_BOTTOM_OPEN = "bottom_open"
     MALFUNCTION_ID_LASERHEADUNIT_MISSING = "laserheadunit_missing"
@@ -35,7 +61,7 @@ class HwMalfunctionHandler(object):
 
         self._messages_to_show = {}
         self._timer = None
-        self.hardware_malfunction = False
+        self._hardware_malfunction = False
 
         self._event_bus.subscribe(
             MrBeamEvents.MRB_PLUGIN_INITIALIZED, self._on_mrbeam_plugin_initialized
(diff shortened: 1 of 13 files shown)


octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1640,11 +1641,49 @@ def _gcode_command_for_cmd(self, cmd):
...
+        Args:
+            current_state: current state
+            new_state: new state
+
+        Returns:
+            None
+        """
+        # if we're unexpected exiting the printing or paused state to locked show the job cancelled  error notification
+        if (
+            current_state == self.STATE_PRINTING and new_state == self.STATE_LOCKED
```
ahmed-mrbeam:
```suggestion
            (current_state == self.STATE_PRINTING or current_state == self.STATE_PAUSED ) and new_state == self.STATE_LOCKED
```

octoprint_mrbeam/static/js/app/view-models/user-notification.js
```diff
@@ -143,7 +176,14 @@ $(function () {
             if ("knowledgebase" in pn_obj) {
                 pn_obj.text += self._getKnowledgeBaseLink(pn_obj.knowledgebase);
             }
+
+            // if error code is present show this instead show error message
             if (
+                notification_conf.err_code &&
+                notification_conf.err_code.length > 0
```
ahmed-mrbeam:
```suggestion
              if (
                   notification_conf.err_code.length 
                   )
```

ahmed-mrbeam (dismissed):

octoprint_mrbeam/static/js/app/view-models/user-notification.js
JFSolution: this is needed to not run in an error

octoprint_mrbeam/iobeam/hw_malfunction_handler.py
```diff
@@ -52,21 +78,42 @@ def report_hw_malfunction_from_plugin(self, malfunction_id, msg, payload=None):
...
+            self._logger.debug("error already reported: %s", malfunction_id)
+
+    def error_already_reported(self, malfunction_id):
+        return self._messages_to_show.get(malfunction_id) is not None
+
+    @property
+    def hardware_malfunction(self):
+        if (
+            self._messages_to_show is None
+            or len(self._messages_to_show) == 0
```
irlaec: I think we should check for `not self._messages_to_show` before this, right?
And I think `self._messages_to_show == {}` would not be reached, as the length is already 0

octoprint_mrbeam/iobeam/iobeam_handler.py
```diff
@@ -1128,11 +1154,21 @@ def _handle_error_message(self, message):
         :param message:
         :return:
         """
-        # TODO: A better way to extract?
```
irlaec, line 1131: Great to get rid of a todo :D

octoprint_mrbeam/iobeam/iobeam_handler.py
```diff
@@ -263,6 +264,18 @@ def send_analytics_request(self, *args, **kwargs):
         """
         return self._send_command(self.get_request_msg([self.DATASET_ANALYTICS]))
 
+    def send_malfunction_request(self, *args, **kwargs):
```
irlaec: I find the name a bit confusing, like we are requesting a malfunction :sweat_smile: 
What do you think about renaming it to something like `check_for_malfunctions` or `send_malfunctions_present_request`

octoprint_mrbeam/mrb_logger.py
```diff
@@ -90,6 +90,9 @@ def log(self, level, msg, *args, **kwargs):
         except TypeError:
             # If it's already unicode we get this TypeError
             pass
+        except Exception as exc:
+            self.log(logging.ERROR, "Error in MrbLogger.log: %s - %s", msg, exc)
+            pass
```
irlaec: We don't need the pass anymore
```suggestion
```

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1640,11 +1641,49 @@ def _gcode_command_for_cmd(self, cmd):
 
         return gcode.group(1)
 
+    def _onExitState(self, current_state, new_state):
```
irlaec:
```suggestion
    def _on_exit_state(self, current_state, new_state):
```

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -1640,11 +1641,49 @@ def _gcode_command_for_cmd(self, cmd):
...
+    def _show_job_cancelled_due_to_internal_error(self):
+        """
+        Shows the job cancelled due to internal error notification.
+
+        Returns:
+            None
+        """
+        notification = (
+            _mrbeam_plugin_implementation.user_notification_system.get_notification(
+                notification_id="job_cancelled_due_to_internal_error",
```
irlaec: Maybe it makes sense to have all the notification ids and error codes in a constants file? I don't know if we have more of these...

irlaec (changes requested):

octoprint_mrbeam/iobeam/hw_malfunction_handler.py
JFSolution: You are right it will not be reached. I check not self._messages_to_show is enough as a single condition

octoprint_mrbeam/iobeam/iobeam_handler.py
JFSolution: I will rename it to `request_available_malfunctions` what you think about this name?

irlaec: Yes, even better :)

octoprint_mrbeam/printing/comm_acc2.py
JFSolution: I like the idea, I will create two new files

irlaec (approved):

PR #1730, opened 2023-03-28
SW-3054 mvp fire detection

JFSolution:
- add high temperature warning
- add high temperature handling in analytics, printer, leds, compressor and exhaust


JFSolution:
Sound is missing but maybe you can have a first look

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -972,6 +984,30 @@ def _event_print_cancelled(self, event, payload, header_extension=None):
         self._add_collector_details()
         self._add_cpu_data(dur=payload.get("time", 0.0))
 
+    def _on_event_laser_job_aborted(self, event, payload, header_extension=None):
+        """Action of a aborted job. Will add a event to analytics and upload it directly.
```
ahmed-mrbeam:
```suggestion
        """Callback for aborted laser job event . Will add an event to analytics and upload it directly.
```

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -1062,6 +1098,59 @@ def _add_other_plugin_data(self, event, event_payload, header_extension=None):
                 "Exception during _add_other_plugin_data: {}".format(e)
             )
 
+    def _on_event_high_temperature_warning(self, event, payload, header_extension=None):
+        """Action of a high temperature warning. Will add a event to analytics with the current temperature and the threshold to trigger the warning.
```
ahmed-mrbeam:
```suggestion
        """Callback for  high temperature warning event. Will add an event to analytics with the current temperature and the set threshold for triggering the user warning.
```

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -1062,6 +1098,59 @@ def _add_other_plugin_data(self, event, event_payload, header_extension=None):
...
+
+        Returns:
+            None
+        """
+        _ = event
+        try:
+            self._add_device_event(
+                AnalyticsKeys.Device.HighTempWarning.TRIGGERED,
+                payload=dict(
+                    temperature=payload.get("tmp", 0),
```
ahmed-mrbeam, line 1112: why 0 if tmp is not there ?, it will not make sense to save an invalid analytics entry

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -1062,6 +1098,59 @@ def _add_other_plugin_data(self, event, event_payload, header_extension=None):
...
+            )
+        except Exception as e:
+            self._logger.exception(
+                "Exception during on_event_high_temperature_warning: {}".format(e)
+            )
+
+    def _on_event_high_temperature_warning_dismissed(
+        self, event, payload, header_extension=None
+    ):
+        """Action of a high temperature warning dismissed. Will add a event to analytics."
```
ahmed-mrbeam:
```suggestion
        """Callback for high temperature warning dismissed event. Will add an event to analytics."
```

octoprint_mrbeam/iobeam/temperature_manager.py
```diff
@@ -89,8 +102,17 @@ def shutdown(self):
         self._shutting_down = True
 
     def reset(self, kwargs):
```
ahmed-mrbeam, line 104: Please add doc string explaining what is the method doing

ahmed-mrbeam (changes requested):

ahmed-mrbeam:
@Josef-MrBeam Please fix the 3 code smells reported by Sonar

octoprint_mrbeam/analytics/analytics_handler.py
JFSolution: We still have the event saved what is the main thing we need the value is just nice to have

JFSolution:
> @Josef-MrBeam Please fix the 3 code smells reported by Sonar

The last remaining codesmell was not introduced with this code change, to get rid of it I would refactor the whole led events file so I will ignore it for now

octoprint_mrbeam/analytics/analytics_handler.py
```diff
@@ -964,6 +976,30 @@ def _event_print_cancelled(self, event, payload, header_extension=None):
         self._add_collector_details()
         self._add_cpu_data(dur=payload["time"])
 
+    def _on_event_laser_job_aborted(self, event, payload, header_extension=None):
```
irlaec: If it's not too much to change, I think it is better to keep it consistent and use `_event_laser_job_finished`.
Then the additional data about the abort (in this case the trigger) would be part of the event `p_aborted`. Here the key for the event is present but not used. 
The structure of the job analytics that we were using so far is ilustrated [here](https://docs.google.com/document/d/1G2UETo65dsFy_X6OhoSsyo0XIFSgjfoFy-n-5RSYJPg/edit#heading=h.wm6ev5f7wbbb)

octoprint_mrbeam/iobeam/laserhead_handler.py
```diff
@@ -496,6 +497,40 @@ def default_laserhead_max_temperature(self):
...
+        # Handle the exceptions
+        if ((isinstance(current_laserhead_properties, dict) is False) or
+                ("high_temperature_offset" not in current_laserhead_properties) or
+                (isinstance(current_laserhead_properties["high_temperature_offset"], float) is False)):
+            # Apply fallback
+            self._logger.debug("Current laserhead properties: {}".format(current_laserhead_properties))
+            self._logger.exception(
+                "Current Laserhead high temperature warn offset couldn't be retrieved, fallback to the temperature value of: {}".format(
+                    self.default_laserhead_high_temperature_warn_offset))
+            return self.default_laserhead_high_temperature_warn_offset
```
irlaec, line 518: I think this is missing the parenthesis. On the other hand, are we planning to add some more logic to `default_laserhead_high_temperature_warn_offset`? Or should we just use the const?

octoprint_mrbeam/iobeam/temperature_manager.py
```diff
@@ -213,7 +247,32 @@ def _stop_if_temp_is_not_current(self):
...
+        """Dismisses the high temperature warning.
+
+        Returns:
+            None
+        """
+        self.high_temperature_warning = False
+        self._event_bus.fire(MrBeamEvents.HIGH_TEMPERATURE_WARNING_DISMISSED)
+
     def _check_temp_val(self):
+        if self.temperature is not None:
```
irlaec: Would it make sense to merge these two ifs?

octoprint_mrbeam/printing/printer.py
```diff
@@ -113,6 +117,20 @@ def fail_print(self, error_msg=None):
...
+        Args:
+            event: event that triggered the abort
+
+        Returns:
+            None
+        """
+        self._comm.abort_print()
+        time.sleep(0.5)
+        self.home(axes="wtf")
+        eventManager().fire(MrBeamEvents.LASER_JOB_ABORTED, dict(trigger=event))
```
irlaec: In the case of cancelled and failed we fire the event from comm_acc2, do you think we should do the same in this case?

irlaec (changes requested):

octoprint_mrbeam/iobeam/laserhead_handler.py
JFSolution: it is a property thats why we don't need the parantheses. Currently we only use the const, I reused the way that ahmed implemented for the values from the start of the laserhead profile.

octoprint_mrbeam/printing/printer.py
JFSolution: Yes, I can move, it was triggering a different event here before, thats why it is still here 😅

octoprint_mrbeam/analytics/analytics_handler.py
JFSolution: But then it would be failed, right?

irlaec (approved):

octoprint_mrbeam/iobeam/laserhead_handler.py
irlaec: Aaah you're right, I didn't notice about the property :)

PR #1745, opened 2023-05-03
SW-3175 fix python statemachine not installed

JFSolution:
- update script will not ignore dependencies
- update script will not force reinstall as this would reinstall all dependencies as well

### DIFF
--- a/octoprint_mrbeam/scripts/update_script.py
+++ b/octoprint_mrbeam/scripts/update_script.py
@@ -179,6 +179,7 @@ def build_wheels(build_queue):
             "--wheel-dir={}".format(
                 tmp_folder
             ),  # Build wheels into <dir>, where the default is the current working directory.
+            "--no-dependencies",  # Don't install package dependencies.
         ]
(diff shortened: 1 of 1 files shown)


octoprint_mrbeam/scripts/update_script.py
```diff
@@ -212,10 +214,10 @@ def install_wheels(install_queue):
             "install",
             "--disable-pip-version-check",
             "--upgrade",  # Upgrade all specified packages to the newest available version. The handling of dependencies depends on the upgrade-strategy used.
-            "--force-reinstall",  # Reinstall all packages even if they are already up-to-date.
```
irlaec, line 215: Could it happen that a package gets damaged and we can't reinstall?

JFSolution: I checked and it seemed that we needed this to install a v0.13.0 for example over the developer 0+unknown version

khaledsherkawi (approved):

PR #1756, opened 2023-05-25
SW-3113 change text of hardware malfunctions

JFSolution:
change text
add new errors
solve SW-3095


octoprint_mrbeam/__init__.py
```diff
@@ -2319,7 +2319,7 @@ def handle_hardware_error_request(self, data):
         Returns:
             NO_CONTENT
         """
-        self.iobeam.send_malfunction_request()
+        self.iobeam.request_available_malfunctions()
```
irlaec, line 2322: Much better :)

octoprint_mrbeam/iobeam/compressor_handler.py
```diff
@@ -144,11 +144,10 @@ def _handle_static_data(self, payload):
                     dataset,
                 )
         else:
-            # If the dataset is empty but we know there is a compressor, something is wrong
-            if self.has_compressor():
-                self._hw_malfunction_handler.report_hw_malfunction_from_plugin(
-                    malfunction_id="compressor", msg="no_compressor_static_data"
-                )
+            self._logger.warn(
```
irlaec, line 149: We don't want to check if there is a compressor anymore? Just to make sure

octoprint_mrbeam/iobeam/compressor_handler.py
```diff
@@ -175,7 +174,9 @@ def _handle_dynamic_data(self, payload):
                         # avoid overheating an potential further damage
                         self.set_compressor_off()
                         self._hw_malfunction_handler.report_hw_malfunction_from_plugin(
-                            malfunction_id="compressor", msg="compressor_rpm_0"
+                            malfunction_id="err_compressor_malfunction",
```
irlaec: Maybe it's a good idea to have all the malfunction ids, messages and error codes in the same place? Then the next time we need to change something they are easy to find. What do you think?

octoprint_mrbeam/iobeam/hw_malfunction_handler.py
```diff
@@ -98,6 +105,8 @@ def report_hw_malfunction(self, dataset, from_plugin=False):
         for malfunction_id, data in dataset.items():
             data = data or {}
             msg = data.get("msg", malfunction_id)
+            # if malfunction_id == "compressor_malfunction":
```
irlaec: Please remove this if not needed

octoprint_mrbeam/iobeam/hw_malfunction_handler.py
```diff
@@ -136,14 +145,55 @@ def show_hw_malfunction_notification(self):
...
+                )
+            elif malfunction_id == self.HW_MANIPULATION:
+                notifications.append(
+                    self._user_notification_system.get_notification(
+                        notification_id="err_interlock_malfunction",
+                        err_code=malfunction.error_code,
+                        replay=True,
+                    )
+                )
+            elif malfunction_id == "err_fan_not_spinning":
```
irlaec: Please save these ids as constants, similar to the previous errors

octoprint_mrbeam/mrb_logger.py
```diff
@@ -86,7 +85,12 @@ def log(self, level, msg, *args, **kwargs):
...
         try:
-            msg = unicode(msg, "utf-8")
+            if isinstance(msg, unicode):
+                # If it's already Unicode, no decoding is needed
+                pass
+            else:
+                # If it's a byte string, convert it to Unicode using "utf-8" encoding
+                msg = unicode(msg, "utf-8")
         except TypeError:
             # If it's already unicode we get this TypeError
```
irlaec, line 95: Now that we already check for unicode, maybe we could add a warn here to see if anything weird happens

octoprint_mrbeam/static/css/mrbeam.css
```diff
@@ -1763,7 +1763,7 @@ ul.dropdown-menu li a {
 
 .ui-pnotify .alert-error a,
 .ui-pnotify .alert-danger a {
-  color: #b94a48;
+  color: #e25303;
```
irlaec, line 1766: I missed that :sweat_smile:

octoprint_mrbeam/static/js/app/view-models/feedback-widget.js
```diff
@@ -62,13 +62,22 @@ $(function () {
             } else {
                 channel = "Stable";
             }
+            if (MRBEAM_LASER_HEAD_MODEL === "x") {
```
irlaec: I've seen we have similar code in convert.js. Maybe it's a good idea to add the "x", "S", "0" in mrbeam.js so that we can access it from everywhere in the frontend. We already have something for the ids:
`mrbeam.laserheadModel = {
    S: 1,
    X: 3,
};`

octoprint_mrbeam/static/js/app/view-models/user-notification.js
```diff
@@ -1,44 +1,152 @@
...
+                        "The device is currently not able to start any laser job.\n" +
+                            "\n" +
+                            "Only disconnect or connect the laserhead unit if device is not powered on!\n" +
+                            "\n" +
+                            "Please shutdown the device and check the connection to the laserhead following this %(opening_tag)sKnowledgebase article%(closing_tag)s.\n" +
+                            "\n" +
+                            "If this doesn’t solve the problem, please %(opening_tag_support)sopen a support ticket%(closing_tag_support)s."
+                    ),
+                    self._getKowledgebaseArticleLinkSprintf(
+                        gettext("https://support.mr-beam.org/en") +
```
irlaec, line 58: Is it enough with this gettext, or is it possible that the rest of the url is different too?

octoprint_mrbeam/static/js/app/view-models/user-notification.js
```diff
@@ -1,44 +1,152 @@
...
+                title: gettext("One Button malfunction"),
+                text: _.sprintf(
+                    gettext(
+                        "The device is currently not able to start any laser job.\n" +
+                            "\n" +
+                            "You can still shut down the device via the Mr Beam Software.\n" +
+                            "\n" +
+                            "Please %(opening_tag_support)sopen a support ticket%(closing_tag_support)s."
+                    ),
+                    self._getKowledgebaseArticleLinkSprintf(self.err)
```
irlaec: I think we need to pass an empty url, before the error code
```suggestion
                    self._getKowledgebaseArticleLinkSprintf(None,self.err)
```

irlaec (changes requested):

octoprint_mrbeam/__init__.py
JFSolution: This was already cahnged but I missed to change the use as well 😅

octoprint_mrbeam/iobeam/compressor_handler.py
JFSolution: They shouldn't change, but I like the idea that you can see them all in one place

octoprint_mrbeam/iobeam/hw_malfunction_handler.py
JFSolution: ohh yes I wanted to do it later and then I forgot

octoprint_mrbeam/mrb_logger.py
JFSolution: I don't see a benefit there, this is just to be able to log something that is already unicode

octoprint_mrbeam/static/css/mrbeam.css
JFSolution: It was not mentioned in the other ticket :)

octoprint_mrbeam/static/js/app/view-models/user-notification.js
JFSolution: The rest is the same, I can remove the text from them as the only thing interesting is the id

octoprint_mrbeam/static/js/app/view-models/user-notification.js
JFSolution: I removed the error as this not needed anymore, I also renamed the function

octoprint_mrbeam/iobeam/compressor_handler.py
JFSolution: This should be fine as we have the 0 rpm error still and the errorcodes from the selftest. I kept it as a log to see if we would need to show a error as well

irlaec (approved):

irlaec (approved):

PR #1757, opened 2023-05-31
Sw 3289 include jsclipper snap plugin into the frontend

hungerpirat:
added js clipper snap plugin which allows:
* easy clipping (union, difference, xor, intersect): elementA.unionClip(elementB) 
* cropping a path to boundaries: elementA.crop(l, b, r, t) 
* path offsetting: elementA.pathOffset(10)


JFSolution:
https://github.com/mrbeam/mrb3-core-plugin/pull/98

PR #1761, opened 2023-06-14
SW-3357 camera logs exeption on camera error and trys over and over again

JFSolution:
fix still showing errors in log but not on frontend

### DIFF
--- a/octoprint_mrbeam/iobeam/lid_handler.py
+++ b/octoprint_mrbeam/iobeam/lid_handler.py
@@ -908,7 +908,7 @@ class PhotoCreator(object):
             if self._plugin.lid_handler.lensCalibrationStarted:
                 # Do not try to do anything fancy during the lens calibration (computationaly lighter)
                 self._logger.debug(
-                    "Lens calibraton - wait for the next picture to take"
+                    "Lens calibration - wait for the next picture to take"
                 )
                 while self._plugin.lid_handler.lensCalibrationStarted:
                     if self.forceNewPic.wait(0.05):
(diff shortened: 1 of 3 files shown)


octoprint_mrbeam/iobeam/lid_handler.py
```diff
@@ -921,6 +921,18 @@ def serve_pictures(
...
             else:
+                if cam.camera_error:
+                    self.stopEvent.set()
+                    self._logger.error(
+                        "Camera error: %s, disabling camera", cam.camera_error
+                    )
+                    self._plugin.user_notification_system.show_notifications(
+                        self._plugin.user_notification_system.get_notification(
+                            "err_cam_conn_err",
+                            err_msg=[str("camera error")],
```
irlaec: Is this str() necessary? Or can we just skip it? 
```suggestion
                            err_msg=["camera error"],
```

octoprint_mrbeam/iobeam/lid_handler.py
```diff
@@ -921,6 +921,18 @@ def serve_pictures(
...
+                            err_code="E-01FF-1006",
```
irlaec: I think there was a module for the error codes, no?

irlaec (changes requested):

octoprint_mrbeam/iobeam/lid_handler.py
JFSolution: you are right :)

octoprint_mrbeam/iobeam/lid_handler.py
JFSolution: ohh I can skip it

irlaec (approved):

PR #1781, opened 2023-07-03
SW-3462 the high temperature abort can be triggered without cooling

JFSolution:
image

Unittests are executed locally and are successful

- raised high tenmperature offset to 3 
- only trigger critical after a second reading after the cooling was triggered


.github/workflows/main.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    uses: mrbeam/mrb-workflows/.github/workflows/build.yml@main
+    uses: mrbeam/mrb-workflows/.github/workflows/build.yml@python2.7
```
khaledsherkawi, line 13: shouldn't this be merged back to "main" once approved?

octoprint_mrbeam/iobeam/temperature_manager.py
```diff
@@ -367,6 +367,7 @@ def _check_temp_val(self):
         if (
             not self.is_cooling(time_wise_only=True)
             and self.temperature > self.temperature_max
+            and self._one_button_handler.is_laser_busy()
```
khaledsherkawi: why not `self._printer.is_busy()`?

.github/workflows/main.yml
JFSolution: No, this will only take affect for the MrBeamPlugin as this is the only one with python2.7 were we still have some developments.

octoprint_mrbeam/iobeam/temperature_manager.py
JFSolution: because I don't have the printer object here 😅

tests/iobeam/test_temperature_manager.py
```diff
@@ -10,12 +10,14 @@
...
 @pytest.fixture()
 def temperature_manager(mrbeam_plugin):
-    temperature_manager = TemperatureManager(mrbeam_plugin)
+    laser = MagicMock()
+    temperature_manager = TemperatureManager(mrbeam_plugin, laser)
     temperature_manager.temperature_max = 50.0
     temperature_manager._high_tmp_warn_offset = 5.0
     temperature_manager._event_bus.fire = MagicMock()
     temperature_manager._analytics_handler = MagicMock()
     temperature_manager._one_button_handler = MagicMock()
```
khaledsherkawi, line 19: do we still need this here or can we user the `laser` to check `is_printing`?

khaledsherkawi: it is used elsewhere in the temp manager so let's keep this

khaledsherkawi (approved):

PR #1793, opened 2023-08-14
SW-3644 update material settings  and link in the MrBeamPlugin and translate

khaledsherkawi:
SW-3644 Update material settings


octoprint_mrbeam/static/js/app/view-models/user-notification.js
```diff
@@ -156,7 +156,7 @@ $(function () {
                     ),
                     self._getPlaceHolderParameters(
                         gettext("https://support.mr-beam.org/en") +
-                            "/support/solutions/articles/43000703002"
+                            "/support/solutions/articles/43000708776-interlock-malfunction-check-the-lid-switch"
```
irlaec, line 159: I think it should be like this instead
```suggestion
                            "/support/solutions/articles/43000708776"
```

octoprint_mrbeam/camera/corners.py
```diff
@@ -25,7 +25,7 @@
 _logger = mrb_logger("octoprint.plugins.mrbeam.camera.corners")
 
 # @logtime()
-def warp_image_by_corners(image, corners, zoomed_out=False):
+def warpImgByCorners(image, corners, zoomed_out=False):
```
irlaec, line 29: Why did we change the name here? It doesn't follow the convention...

irlaec (changes requested):

octoprint_mrbeam/static/js/app/view-models/user-notification.js
khaledsherkawi: This is a completely different article and it was a request to change it

octoprint_mrbeam/camera/corners.py
khaledsherkawi: We changed this in a refactor a while back and I just found out that this is used by the camera plugin and the plugin is failing to load because of it so I reverted the name

irlaec: Aaaah okay! Could you please add a comment so we don't do it again?

khaledsherkawi: done

irlaec (approved):

PR #1796, opened 2023-08-31
SW-3824 Revert renaming variable to get along with camera plugin

caglarkilimci:


khaledsherkawi (approved):

khaledsherkawi (changes requested):
Please add a comment for all methods being used by the camera plugin. Something like:
# This method is being used by the camera plugin
# Do not modify without checking the usage in the camera plugin

amrithvenkat (approved):

PR #1800, opened 2023-09-18
SW-3537 laserhead none will crash plugin

JFSolution:
prevent recursive log calls

### DIFF
--- a/octoprint_mrbeam/mrb_logger.py
+++ b/octoprint_mrbeam/mrb_logger.py
@@ -34,6 +34,8 @@ class MrbLogger(object):
         self.id = id
         self.id_short = self._shorten_id(id)
         self.my_buffer = []
+        self.messages_to_log = []  # list of messages to log, to prevent recursive logs
+        self.recursive_depth = 0
         # TODO: this line overrides logging.yaml!!!
         if lvl is not None:
             self.logger.setLevel(lvl)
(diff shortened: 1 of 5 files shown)


octoprint_mrbeam/mrb_logger.py
```diff
@@ -84,6 +86,17 @@ def log(self, level, msg, *args, **kwargs):
         :param terminal_dump: Collect and log a terminal dump. Terminal dumps are also sent to analytics if analytics is not explicitly set to False.
         :type kwargs:
         """
+        if "Recursive call for log: %s" % msg in self.messages_to_log:
```
irlaec: Shall we save `"Recursive call for log: %s"` in a constant, to reuse it later?

octoprint_mrbeam/mrb_logger.py
```diff
@@ -96,7 +109,10 @@ def log(self, level, msg, *args, **kwargs):
             # If it's already unicode we get this TypeError
             pass
         except Exception as exc:
-            self.log(logging.ERROR, "Error in MrbLogger.log: %s - %s", msg, exc)
+            # self.log(logging.ERROR, "Error in MrbLogger.log: %s - %s", msg, exc)
```
irlaec: If we don't need it, please remove this line

irlaec (changes requested):

irlaec (dismissed):

irlaec (approved):

PR #1801, opened 2023-09-18
SW-3537 improve recursive log prevention

JFSolution:


hungerpirat (approved):
At some point we should limit the maximum amount of entries in the self.messages_to_log dict. 
@Josef-MrBeam knows.

PR #1811, opened 2023-09-22
SW-2304 frontend implement air filter iii details in the settings maintenance page

JFSolution:
detect AF change and update usage info in frontend

### DIFF
--- a/octoprint_mrbeam/__init__.py
+++ b/octoprint_mrbeam/__init__.py
@@ -514,7 +514,7 @@ class MrBeamPlugin(
         )
 
     def on_settings_load(self):
-        self._send_maintenance_information(trigger="settings_load")
+        self._send_maintenance_information(trigger=MrBeamEvents.SETTINGS_LOAD)
         return dict(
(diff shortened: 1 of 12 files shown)


octoprint_mrbeam/__init__.py
```diff
@@ -683,6 +696,8 @@ def on_settings_save(self, data):
             octoprint.plugin.SettingsPlugin.on_settings_save(self, data)
 
     def set_heavy_duty_prefilter(self, value):
+        if value != self._settings.get(["heavyDutyPrefilter"]):
+            self._send_maintenance_information(trigger="heavyDutyPrefilter_changed")
```
irlaec: Maybe it makes sense to group the different triggers in an enum or similar?

irlaec (approved):

octoprint_mrbeam/__init__.py
JFSolution: ahh sure, I added them for debugging but it makes sense to keep them and change them to enums or constants

irlaec (approved):

PR #1814, opened 2023-09-28
SW-3812 Frontend - Refresh the web page if the laser cutter mode flips

irlaec:


octoprint_mrbeam/static/js/app/view-models/mother.js
```diff
@@ -537,6 +549,8 @@ $(function () {
                         MRBEAM_PLUGIN_VERSION +
                         ", isFirstRun: " +
                         CONFIG_FIRST_RUN +
+                        ", backendLaserCutterMode: frontend=" +
```
khaledsherkawi: since this whole section is just frontend checks and it is mentioned at the top
```suggestion
                        ", backendLaserCutterMode: " +
```

khaledsherkawi (approved):

octoprint_mrbeam/static/js/app/view-models/mother.js
irlaec: Oh the log didn't make much sense... changed it :sweat_smile:

PR #1815, opened 2023-09-28
SW-3809 - Backend- Select materials csv based on laser cutter mode

irlaec:


octoprint_mrbeam/util/material_csv_parser.py
```diff
@@ -16,6 +16,15 @@
...
 VALID_MODELS = [MRBEAM, MRB_DREAMCUT, MRB_DREAMCUT_S, MRB_READY, MRB_DREAMCUT_X]
 
+DEFAULT_MODE = "default"
+ROTARY_MODE = "rotary"
+
+MATERIALS_CSV_DIR = "files/material_settings/"
+MATERIALS_CSVS = {
+    DEFAULT_MODE: MATERIALS_CSV_DIR + "materials.csv",
+    ROTARY_MODE: MATERIALS_CSV_DIR + "materials_rotary.csv",
+}
```
khaledsherkawi:
```suggestion
MATERIALS_CSV_DIR = "files/material_settings/"
DEFAULT_MATERIALS_CSV_DIR = MATERIALS_CSV_DIR + "materials.csv"
ROTARY_MATERIALS_CSV_DIR = MATERIALS_CSV_DIR + "materials_rotary.csv"
MATERIALS_CSVS = {
    DEFAULT_MODE: DEFAULT_MATERIALS_CSV_DIR,
    ROTARY_MODE: ROTARY_MATERIALS_CSV_DIR,
}
```

octoprint_mrbeam/util/material_csv_parser.py
```diff
@@ -16,6 +16,15 @@
 
 VALID_MODELS = [MRBEAM, MRB_DREAMCUT, MRB_DREAMCUT_S, MRB_READY, MRB_DREAMCUT_X]
 
+DEFAULT_MODE = "default"
+ROTARY_MODE = "rotary"
```
khaledsherkawi, line 21:
```suggestion
# TODO: import these from mode services
DEFAULT_MODE = "default"
ROTARY_MODE = "rotary"
```

octoprint_mrbeam/materials.py
```diff
@@ -1,12 +1,16 @@
...
-from octoprint_mrbeam.mrb_logger import mrb_logger
-from octoprint_mrbeam.util import dict_get
 
+from octoprint_mrbeam.model.custom_material_model import CustomMaterialModel
+from octoprint_mrbeam.mrb_logger import mrb_logger
 
 # singleton
 _instance = None
 
+# TODO: SW-3719 keep these constants in a unique place
```
khaledsherkawi:
```suggestion
# TODO: SW-3719 import these from mode services
```

octoprint_mrbeam/__init__.py
```diff
@@ -1282,16 +1274,17 @@ def custom_materials(self, data):
...
-            if "delete" in data:
-                materials(self).delete_custom_material(data["delete"])
+            if data.get("delete", []):
+                for material_key in data["delete"]:
+                    materials(self).delete_custom_material(material_key)
 
-            if "put" in data and isinstance(data["put"], dict):
-                for key, m in data["put"].iteritems():
-                    materials(self).put_custom_material(key, m)
+            if isinstance(data.get("put"), dict):
```
khaledsherkawi, line 1283: This was it has a fall back that is equivalent to (if "put" in data)
```suggestion
            if isinstance(data.get("put", None), dict):
```

octoprint_mrbeam/__init__.py
```diff
@@ -1300,6 +1293,16 @@ def custom_materials(self, data):
...
 
+    def get_laser_cutter_mode(self):
+        # TODO: SW-3719 return actual laser cutter mode
+        return "default"
+
+    def get_laser_head_model(self):
+        return self.laserhead_handler.get_current_used_lh_data()["model"]
+
+    def get_laser_head_serial(self):
+        return self.laserhead_handler.get_current_used_lh_data()["serial"]
```
khaledsherkawi: I think these 2 methods should be get_current_laser_head... as there is more than 1 laser head model and serial (current and old)

octoprint_mrbeam/materials.py
```diff
@@ -26,122 +29,24 @@ def __init__(self, plugin):
...
-        return res
+        return custom_materials
 
-    def reset_all_custom_materials(self):
-        self._logger.info("Resetting all custom material settings!!!!")
-        self.custom_materials = {}
-        self._save(force=True)
-
-    def _load(self, force=False):
-        if not self.custom_materials_loaded or force:
```
khaledsherkawi, line 110: why aren't we using this property anymore? shall we remove it?

octoprint_mrbeam/materials.py
```diff
@@ -152,17 +57,46 @@ def _save(self, force=False):
...
-            )
-        except:
-            self._logger.exception(
-                "Exception while writing custom materials to file %s",
-                self.custom_materials_file,
+            self._logger.info("{} custom materials saved".format(len(self.custom_materials)))
+        except Exception as e:
+            self._logger.exception("Exception while saving custom materials: {}".format(e))
+
+    def get_custom_materials_for_laser_cutter_mode(self):
```
khaledsherkawi, line 64: Since the method doesn't have any parameters and just gets the custom material (with the current mode as it does with the current laser head) then I would not have it name "for_laser_cutter_mode"

khaledsherkawi: The laser head definition is still in the frontend but we will port it as some point to the backend and it should have only 1 method to get custom materials regardless of the underlaying current conditions

khaledsherkawi: maybe something like get_custom_materials_by_settings, more general

octoprint_mrbeam/materials.py
```diff
@@ -1,12 +1,16 @@
 import os
```
khaledsherkawi, line 1: Love the refactor 😍

octoprint_mrbeam/model/custom_material_model.py
```diff
@@ -0,0 +1,57 @@
+
```
khaledsherkawi, line 1: Let's talk about this in a call

khaledsherkawi (changes requested):

octoprint_mrbeam/__init__.py
irlaec: Actually both are the same, because the default value is None: `dict.get(key, default = None)`

octoprint_mrbeam/__init__.py
irlaec: Agree, I'll change it

octoprint_mrbeam/materials.py
irlaec: You mean self.custom_materials_loaded? Or force?
custom_materials_loaded isn't necessary because we load them in the __init__
and force was to read them from the file again, but in the current implementation it isn't necessary.
Both are removed in the current code

tests/test_materials.py
```diff
@@ -0,0 +1,71 @@
+from mock.mock import MagicMock
```
khaledsherkawi, line 1: Optional: Add 'get' test for an empty custom materials dict

khaledsherkawi (approved):

tests/test_materials.py
irlaec: Done

PR #1819, opened 2023-10-05
SW-3882 implement the usage percentage for the new filters

JFSolution:
*add logic to calculate usage with pressure for af3
* write usage data on pressure update
* implement pressure save logic
* remove debug logging
* clean up about view model
* for af3 don't use 10 percent hops
* fix settings not loaded correctly bug

### DIFF
--- a/octoprint_mrbeam/analytics/usage_handler.py
+++ b/octoprint_mrbeam/analytics/usage_handler.py
@@ -1134,7 +1134,10 @@ class UsageHandler(object):
                 self._logger.info(
                     "Pressure value is higher than the last saved pressure value. Updating the pressure value."
                 )
-                self._update_pressure_value(pressure, stage_key)
+                new_pressure_value = (
+                    last_saved_pressure_value + 1
+                )  # don't override just increment the value to prevent jumps
+                self._update_pressure_value(new_pressure_value, stage_key)
             elif pressure < last_saved_pressure_value and percent_difference > 20:
                 self._logger.info(
                     "Pressure drop of 20% detected reset time and pressure value."
--- a/octoprint_mrbeam/analytics/usage_handler.py
+++ b/octoprint_mrbeam/analytics/usage_handler.py
@@ -888,9 +888,14 @@ class UsageHandler(object):
         if self._airfilter is None:
             return -1
         if self._airfilter.model_id in AirFilter.AIRFILTER3_MODELS:
-            return self._calculate_af3_filter_usage(filter_stage=AirFilter.CARBONFILTER)
+            percent = self._calculate_af3_filter_usage(
+                filter_stage=AirFilter.CARBONFILTER
+            )
         else:
-            return self._calculate_af2_filter_usage(filter_stage=AirFilter.CARBONFILTER)
+            percent = self._calculate_af2_filter_usage(
+                filter_stage=AirFilter.CARBONFILTER
+            )
+        return round(percent)
 
     def get_prefilter_usage(self):
         """
@@ -902,9 +907,10 @@ class UsageHandler(object):
         if self._airfilter is None:
             return -1
         if self._airfilter.model_id in AirFilter.AIRFILTER3_MODELS:
-            return self._calculate_af3_filter_usage(filter_stage=AirFilter.PREFILTER)
+            percent = self._calculate_af3_filter_usage(filter_stage=AirFilter.PREFILTER)
         else:
-            return self._calculate_af2_filter_usage(filter_stage=AirFilter.PREFILTER)
+            percent = self._calculate_af2_filter_usage(filter_stage=AirFilter.PREFILTER)
+        return round(percent)
 
     def _calculate_af2_filter_usage(self, filter_stage):
         """
@@ -946,7 +952,7 @@ class UsageHandler(object):
         Returns:
             The usage of the given filter stage in percent.
         """
-        usage_data = None
+        usage_data = {}
         stage_usage_time = None
         pressure_graph = None
 
@@ -1035,6 +1041,23 @@ class UsageHandler(object):
         """
         # Separate x and y values from the known points
         x_values, y_values = zip(*reference_points)
+
+        # limit input value
+        if value > max(x_values):
+            mrb_logger("octoprint.plugins.mrbeam.iobeam.dustmanager").error(
+                "value %s is higher than max value %s, limiting to max value",
+                value,
+                max(x_values),
+            )
+            value = max(x_values)
+
+        elif value < min(x_values):
+            mrb_logger("octoprint.plugins.mrbeam.iobeam.dustmanager").error(
+                "value %s is lower than min value %s, limiting to min value",
+                value,
+                min(x_values),
+            )
+            value = min(x_values)
 
         percentage = float(
             UsageHandler.linear_piecewise_interpolate(value, x_values, y_values)
--- a/octoprint_mrbeam/analytics/usage_handler.py
+++ b/octoprint_mrbeam/analytics/usage_handler.py
@@ -1108,23 +1108,44 @@ class UsageHandler(object):
                 )
                 stage_key = self.CARBON_FILTER_KEY
                 reset_call = self.reset_carbon_filter_usage
+                value_change = min(
+                    abs(last_saved_pressure_value - pressure),
+                    AirFilter.AF3_MAX_CARBON_FILTER_PRESSURE_CHANGE,
+                )  # limit the value change to AirFilter.MAX_CARBON_FILTER_PRESSURE_CHANGE
+                pressure_percentage_new = self.get_precentage_from_interpolation(
+                    AirFilter.AF3_PRESSURE_GRAPH_CARBON_FILTER, pressure
+                )
+                pressure_percentage_last_saved = self.get_precentage_from_interpolation(
+                    AirFilter.AF3_PRESSURE_GRAPH_CARBON_FILTER,
+                    last_saved_pressure_value,
+                )
             elif filter_stage == AirFilter.PREFILTER:
                 last_saved_pressure_value = (
                     self._get_airfilter_prefilter_usage_data().get(self.PRESSURE_KEY, 0)
                 )
                 stage_key = self.PREFILTER_KEY
                 reset_call = self.reset_prefilter_usage
+                value_change = min(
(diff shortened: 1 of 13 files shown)


octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -466,8 +467,8 @@ def write_usage_analytics(self, action=None):
         try:
             usage_data = dict(
                 total=self._usage_data[self.TOTAL_KEY][self.JOB_TIME_KEY],
-                prefilter=self.get_prefilter_usage(),
-                carbon_filter=self.get_carbon_filter_usage,
+                prefilter=self._get_prefilter_usage_time(),
+                carbon_filter=self._get_carbon_filter_usage_time,
```
irlaec: I think parenthesis are missing
```suggestion
                carbon_filter=self._get_carbon_filter_usage_time(),
```

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -880,3 +877,307 @@ def _migrate_old_airfilter_structure_to_new(self):
...
+        Returns:
+            The usage of the given filter stage in percent.
+        """
+        # get the correct data for the filter stage
+        if filter_stage == AirFilter.PREFILTER:
+            stage_usage_time = self._get_prefilter_usage_time()
+        elif filter_stage == AirFilter.CARBONFILTER:
+            stage_usage_time = self._get_carbon_filter_usage_time()
+
+        usage_time = stage_usage_time if stage_usage_time is not None else 0
```
irlaec: Another option (just a suggestion)
```suggestion
        usage_time = stage_usage_time or 0
```

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -880,3 +877,307 @@ def _migrate_old_airfilter_structure_to_new(self):
...
+
+        # calculate the percentage
+        time_percentage = self._get_percentage_from_time(
+            usage_time, self._airfilter.get_lifespans(filter_stage)[0]
+        )
+
+        # calculate the total percentage
+        total_percentage = (
```
irlaec, line 943: Can we just remove the parenthesis?

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -880,3 +877,307 @@ def _migrate_old_airfilter_structure_to_new(self):
...
+        elif filter_stage == AirFilter.CARBONFILTER:
+            pressure_graph = AirFilter.AF3_PRESSURE_GRAPH_PREFILTER
+            usage_data = self._get_airfilter_carbon_filter_usage_data()
+            stage_usage_time = self._get_carbon_filter_usage_time()
+
+        # get the global values
+        pressure_loss = usage_data.get(
+            self.PRESSURE_KEY, AirFilter.MAX_PRESSURE_DIFFERENCE
+        )
+        usage_time = stage_usage_time if stage_usage_time is not None else 0
```
irlaec: Same suggestion
```suggestion
        usage_time = stage_usage_time or 0
```

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -880,3 +877,307 @@ def _migrate_old_airfilter_structure_to_new(self):
...
+
+        Returns:
+            The percentage for the given value.
+        """
+        # Separate x and y values from the known points
+        x_values, y_values = zip(*reference_points)
+
+        # limit input value
+        if value > max(x_values):
+            mrb_logger("octoprint.plugins.mrbeam.iobeam.dustmanager").error(
```
irlaec:
```suggestion
            self._logger.error(
```

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -880,3 +877,307 @@ def _migrate_old_airfilter_structure_to_new(self):
...
+                "value %s is higher than max value %s, limiting to max value",
+                value,
+                max(x_values),
+            )
+            value = max(x_values)
+
+        elif value < min(x_values):
+            mrb_logger("octoprint.plugins.mrbeam.iobeam.dustmanager").error(
```
irlaec:
```suggestion
            self._logger.error(
```

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -880,3 +877,307 @@ def _migrate_old_airfilter_structure_to_new(self):
...
+    def _get_percentage_from_time(usage_time, lifespan):
+        """
+        Calculate the percentage of the lifespan that has been used.
+
+        Args:
+            usage_time: usage time in seconds
+            lifespan: lifespan in hours
+
+        Returns:
+            float: percentage of lifespan that has been used
```
irlaec, line 1082: is it a float? Because for example if usage_time is 0 it would be an int, no?

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -880,3 +877,307 @@ def _migrate_old_airfilter_structure_to_new(self):
...
+                    self.AIRFILTER_KEY,
+                    self._get_airfilter_serial(),
+                    self.CARBON_FILTER_KEY,
+                    self.FAN_TEST_RPM_KEY,
+                ],
+                rpm,
+            )
+            self._write_usage_data()
+
+    def _assign_value_and_create_structure(self, nested_keys, value):
```
irlaec: Could you please add docstrings for this?
Also, is there a reason to use current_dict instead of self._usage_data?
And last, in `current_dict[nested_keys[-1]]`, is nested_keys[-1]] always supposed to be self.FAN_TEST_RPM_KEY? 

Sorry for all the questions :sweat_smile:

octoprint_mrbeam/iobeam/airfilter.py
```diff
@@ -206,6 +237,81 @@ def set_pressure(
         if pressure4 is not None:
             self._pressure4 = pressure4
 
+        if (
```
irlaec: Just a suggestion if you want to try it:
`if all(p is not None for p in [pressure1, pressure2, pressure3, pressure4]):`
or even 
`if all([pressure1, pressure2, pressure3, pressure4]):`
if we want to discard other falsy values as "" or False

octoprint_mrbeam/iobeam/dust_manager.py
```diff
@@ -592,6 +652,7 @@ def _check_if_fan_is_spinning_during_a_running_job(self):
             self._fan_not_spinning_ts = None
 
     def _check_and_report_error(self):
```
irlaec, line 654: Please add docstrings

irlaec (changes requested):

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: doesn't work as this is a static method

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: samer here

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: Ok will add
no, no perticular reasos
no it is the last element of the nested keys can be anything

octoprint_mrbeam/iobeam/dust_manager.py
JFSolution: done

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: good catch, I fixed it

octoprint_mrbeam/analytics/usage_handler.py
irlaec: Do we want to keep it static, or was it just because before we weren't using anything from the class?

octoprint_mrbeam/analytics/usage_handler.py
irlaec: So if I understand correctly the "value" is only assigned to the last nested_key, right? If that's the case, could you please clarify this in the docstring or in the name of the method?

octoprint_mrbeam/analytics/usage_handler.py
```diff
@@ -1168,6 +1169,13 @@ def set_fan_test_rpm(self, rpm):
...
     def _assign_value_and_create_structure(self, nested_keys, value):
+        """
+        Assign the given value to the given nested keys and create the structure if it does not exist.
+
+        Args:
+            nested_keys: The nested keys to assign the value to.
+            value: The value to assign.
+        """
         current_dict = self._usage_data
         for key in nested_keys[:-1]:
```
irlaec: And is there a reason to loop in the opposite direction? Or couple it be simply
`for key in nested_keys`

octoprint_mrbeam/analytics/usage_handler.py
irlaec: This is missing :) 
But actually we could also just return time_percentage directly, as you wish

irlaec (changes requested):

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: The only thing we would use is the logging, but to make it easier I can move it to non static back

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: Yes that is correct, it will create the nested dict and assign the value to the last key

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: it is not the opposit direction, it will exclude the last element

irlaec: Aaaah you're right, sorry I misunderstood :sweat_smile:

octoprint_mrbeam/analytics/usage_handler.py
JFSolution: I changed it but it is autoformated like this.
The idea was here to mention that is currently only the time but we can add more

irlaec (approved):

PR #1824, opened 2023-10-11
SW-2320 update air filter profiles

JFSolution:
update profiles
fix af1
add hose blocked malfunction


octoprint_mrbeam/iobeam/airfilter.py
```diff
@@ -296,6 +307,11 @@ def pressure_drop_mainfilter(self):
             return mainfilter_pressure_avg - fan_pressure_avg
         return None
 
+    def exhaust_hose_is_blocked(self):
+        if self.model_id in self.AIRFILTER3_MODELS:
+            return self._pressure2 < self.AF3_PRESSURE2_MIN
+        return None
```
irlaec, line 313: Would it be more clear to return False?

octoprint_mrbeam/iobeam/airfilter.py
```diff
@@ -337,6 +357,36 @@ def set_temperatures(
         if temperature4 is not None:
             self._temperature4 = temperature4
 
+    def set_connected(self, connected):
```
irlaec: Having used @property before, maybe we can use @setter here?

irlaec: Or maybe not because there's quite a bit of logic...

irlaec (approved):

octoprint_mrbeam/iobeam/airfilter.py
JFSolution: I would say no, as we don't know it for sure, it is a unknown state

octoprint_mrbeam/iobeam/airfilter.py
JFSolution: But a setter can have logic, thats why there is a setter and not direct access, right?

PR #1831, opened 2023-10-22
SW-3951 backend refactor the printer profile class and use it to implement rotary profile

khaledsherkawi:
SW-3951 Add profile class
The profile class is a modified version of the printerProfile class from OctoPrint. It has been modified to be able to serve as a parent class that can be implemented for different purposes and for different profile schemas.

SW-3951 Add laser cutter profile model

SW-3951 Add laser cutter profile service
This class implements the profile parent class to be used for laser cutter profiles.

SW-3951 Add laser cutter profiles to constants

SW-3951 Refactor services

SW-3951 Remove duplicated class

SW-3951 Refactor laser cutter profiles viewmodel

SW-3951 Implement the new laser cutter profile service


octoprint_mrbeam/__init__.py
```diff
@@ -301,8 +296,23 @@ def initialize(self):
...
+    def update_laser_cutter_profile(self, profile):
+        self.laser_cutter_profile_service.save(profile, allow_overwrite=True, make_default=True)
+        self.laser_cutter_profile_service.select(profile['id'])
+
+    def get_laser_cutter_profile_for_current_configuration(self):
+        laser_cutter_mode = self.get_laser_cutter_mode()
+        device_series = self._device_info.get_series()
+        profile = laser_cutter_profiles.default_profile
+        if laser_cutter_mode == LaserCutterModeEnum.DEFAULT.value:
+            if device_series == "2C":
```
irlaec: Do we have the device series somewhere in a constant? Or maybe it's a good moment to create an enum :slightly_smiling_face:

octoprint_mrbeam/__init__.py
```diff
@@ -335,14 +345,29 @@ def _on_laserhead_ready(self, *args, **kwargs):
         self._laserhead_ready = True
         self._try_to_connect_laser()
 
+    def _on_laser_cutter_profile_initialized(self, *args, **kwargs):
```
irlaec, line 372: If we don't need the *args, **kwargs, is it possible to remove them? like
```suggestion
    def _on_laser_cutter_profile_initialized(self, _, _):
```
Or maybe just remove it from the docstring

octoprint_mrbeam/__init__.py
```diff
@@ -335,14 +345,29 @@ def _on_laserhead_ready(self, *args, **kwargs):
...
+        """Called when the laser cutter profile is initialized is initialized.
```
irlaec:
```suggestion
        """Called when the laser cutter profile is initialized.
```

octoprint_mrbeam/__init__.py
```diff
@@ -1700,31 +1725,16 @@ def engraveCalibrationMarkers(self, intensity, feedrate):
     # 		return NO_CONTENT
 
     # Laser cutter profiles
-    @octoprint.plugin.BlueprintPlugin.route("/profiles", methods=["GET"])
+    @octoprint.plugin.BlueprintPlugin.route("/currentProfile", methods=["GET"])
     def laserCutterProfilesList(self):
```
irlaec: I think the name is for the old functionality. This is just a suggestion, just choose one following the convention if we have any:
```suggestion
    def get_current_laser_cutter_profile(self):
```

octoprint_mrbeam/__init__.py
```diff
@@ -2117,7 +2127,7 @@ def on_api_command(self, command, data):
                         parse_csv(
                             device_model=self.get_model_id(),
                             laserhead_model=self.get_current_laser_head_model(),
-                            laser_cutter_mode=self.get_laser_cutter_mode(),
+                            laser_cutter_mode = self.get_laser_cutter_mode() if self.get_laser_cutter_mode() is not None else LaserCutterModeEnum.DEFAULT.value,
```
irlaec: Suggestion (no need to change it if you prefer the other way):
```suggestion
                            laser_cutter_mode = self.get_laser_cutter_mode() or LaserCutterModeEnum.DEFAULT.value,
```

octoprint_mrbeam/constant/profile/laser_cutter/series_2c.py
```diff
@@ -1,12 +1,17 @@
 #!/usr/bin/env python3
 
 profile = dict(
-    id="MrBeam2C",
+    id="series_2c",
```
irlaec: Maybe we could use here the constant/enum I mentioned before

octoprint_mrbeam/constant/profile/laser_cutter/__init__.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import absolute_import
+
+from octoprint.util import dict_merge
+from . import default, series_2c, rotary, series_2c_rotary
```
irlaec: [PEP8](https://peps.python.org/pep-0008/#imports:~:text=Absolute%20imports%20are%20recommended%2C%20as%20they%20are%20usually%20more%20readable%20and%20tend%20to%20be%20better%20behaved%20(or%20at%20least%20give%20better%20error%20messages)%20if%20the%20import%20system%20is%20incorrectly%20configured%20(such%20as%20when%20a%20directory%20inside%20a%20package%20ends%20up%20on%20sys.path)%3A) recommends absolute imports

octoprint_mrbeam/printing/comm_acc2.py
```diff
@@ -164,7 +164,7 @@ def __init__(
         self._port = port
         self._baudrate = baudrate
         self._callback = callbackObject
-        self._laserCutterProfile = laserCutterProfileManager().get_current_or_default()
+        self._laserCutterProfile = printerProfileManager.get_current_or_default()
```
irlaec: In the __init__ the printerProfileManager argument is optional... should we change it? Otherwise we need to handle the case where it is None.
And why do we use the printerProfileManager now? We can discuss it on a call :slightly_smiling_face:

octoprint_mrbeam/printing/printer.py
```diff
@@ -72,8 +74,8 @@ def connect(self, port=None, baudrate=None, profile=None):
         if self._comm is not None:
             self._comm.close()
 
-        eventManager().fire(Events.CONNECTING, payload=dict(profile=profile))
-        self._printerProfileManager.select(profile)
+        eventManager().fire(Events.CONNECTING, payload=dict(profile=self._printerProfileManager.get_current_or_default()['id']))
+        # self._printerProfileManager.select(profile)
```
irlaec, line 81: Please remove this commented out code if we don't need it

octoprint_mrbeam/printing/printer.py
```diff
@@ -48,6 +49,7 @@ def __init__(self, fileManager, analysisQueue, printerProfileManager):
             current_z=None,
         )
         self._user_notification_system = None
+        self._printerProfileManager = laser_cutter_profile_service()
```
irlaec, line 54: Should it be _laser_cutter_profile_service instead, or is this something related to Octoprint? If it's the second, could you add a comment please?

octoprint_mrbeam/printing/printer.py
```diff
@@ -112,7 +131,7 @@ def cancel_print(self):
         """Cancel the current printjob and do homing."""
         super(Laser, self).cancel_print()
         time.sleep(0.5)
-        self.home(axes="wtf")
```
irlaec, line 115: :clap:

octoprint_mrbeam/printing/printer.py
```diff
@@ -92,18 +94,35 @@ def set_colors(self, currentFileName, value):
...
+            after_homing_shift_rate=printer_profile["volume"]["after_homing_shift_rate"],
         )
         self._comm.rescue_from_home_pos()
         command = "G92X{x}Y{y}Z{z}".format(**params)
-        self.commands(["$H", command, "G90", "G21"])
+
+        # Moving command after homing
+        moving_command = "G1X{after_homing_shift_x}Y{after_homing_shift_y}F{after_homing_shift_rate}".format(**params)
+
+        """
```
irlaec: Maybe move this to the docstring? And we can add there the comment about the axes arg too

octoprint_mrbeam/service/profile/laser_cutter_profile.py
```diff
@@ -0,0 +1,56 @@
...
+            dict(),
+        )
+
+    return _instance
+
+
+class LaserCutterProfileService(ProfileService):
+    """ Service class for laser cutter profile. """
+
+    DEFAULT_PROFILE_ID = "default"
```
irlaec: We could use LaserCutterModeEnum.DEFAULT.value here

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+    load and save profiles.
+
+    A profile is a ``dict`` of a certain structure.
+
+    Args:
+        default (dict): The default profile to use if no other profile is selected.
+        folder (str): The folder to store the profiles in.
+    """
+
+    DEFAULT_PROFILE_ID = "_default"
```
irlaec, line 51: And here too (if we can use "default" instead of "_default")

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+
+        default_overrides["id"] = self.DEFAULT_PROFILE_ID
+        self.save(default_overrides)
+
+        settings().set(["profiles", self._id, "defaultProfile"], None, force=True)
+        settings().save()
+
+        self._logger.info("Migrated default profile from settings to {}.profile: {!r}".format(self.DEFAULT_PROFILE_ID, default_overrides))
+
+    def _verify_default_available(self):
```
irlaec, line 93: Let's discuss this in the call too

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+            default_id = self.DEFAULT_PROFILE_ID
+
+        profile = self.get(default_id)
+        if profile is None:
+            self._logger.error("Selected default profile {} is invalid, resetting to default values".format(default_id))
+            profile = copy.deepcopy(self._default)
+            profile["id"] = default_id
+            self.save(self._default, allow_overwrite=True, make_default=True)
+
+    def select(self, identifier):
```
irlaec, line 122: For select, remove, etc: what is the boolean we return? Maybe we can make the names of the methods more self-explanatory, or at least add some docstrings

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+    def save(self, profile, allow_overwrite=False, make_default=False):
+        if "id" in profile:
+            identifier = profile["id"]
+        elif "name" in profile:
+            identifier = profile["name"]
+        else:
+            raise InvalidProfileError("profile must contain either id or name")
+
+        identifier = self._sanitize(identifier)
+        profile["id"] = identifier
```
irlaec, line 164: Maybe up to this point we could have a single method called something like _generate_profile_id(profile)

irlaec: Or continue more, and just _sanitize_profile(profile) where we do everything until _save_to_path

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+
+    @property
+    def last_modified(self):
+        dates = [os.stat(self._folder).st_mtime]
+        dates += [entry.stat().st_mtime for entry in scandir(self._folder) if entry.name.endswith(".profile")]
+        return max(dates)
+
+    def get_default(self):
+        default = settings().get(["profiles", self._id, "default"])
+        if default is not None and self.exists(default):
```
irlaec, line 196: I think this could be simplified, as we already handle the None value inside of self.exists()
```suggestion
        if self.exists(default):
```

irlaec: Or even more, in self.get() we already check if it exists, so we wouldn't need this statement

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+            if profile is not None:
+                return profile
+            else:
+                self._logger.warn("Default profile {} is invalid, falling back to built-in defaults".format(default))
+
+        return copy.deepcopy(self._default)
+
+    def set_default(self, identifier):
+        all_identifiers = self._load_all_identifiers().keys()
+        if identifier is not None and not identifier in all_identifiers:
```
irlaec: I think it can be simplified:
```suggestion
        if identifier not in all_identifiers:
```

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+
+            identifier = entry.name[:-len(".profile")]
+            results[identifier] = entry.path
+        return results
+
+    def _load_from_path(self, path):
+        if not os.path.exists(path) or not os.path.isfile(path):
+            return None
+
+        import yaml
```
irlaec, line 262: Can you please move this import to the top?

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+        with open(path) as f:
+            profile = yaml.safe_load(f)
+
+        if profile is None or not isinstance(profile, dict):
```
irlaec, line 266:
```suggestion
        if not isinstance(profile, dict):

```

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+
+    def _save_to_path(self, path, profile, allow_overwrite=False):
+        validated_profile = self._ensure_valid_profile(profile)
+        if not validated_profile:
+            raise InvalidProfileError()
+
+        if os.path.exists(path) and not allow_overwrite:
+            raise SaveError("Profile %s already exists and not allowed to overwrite" % profile["id"])
+
+        from octoprint.util import atomic_write
```
irlaec, line 292: Same with these imports

octoprint_mrbeam/service/profile/profile.py
```diff
@@ -0,0 +1,482 @@
...
+        sanitized_name = ''.join(c for c in name if c in valid_chars)
+        sanitized_name = sanitized_name.replace(" ", "_")
+        return sanitized_name
+
+    def _migrate_profile(self, profile):
+        """
+        Subclasses must override this method.
+        """
+        return False
+        # raise NotImplementedError("Subclasses must implement _migrate_profile.")
```
irlaec: In the call, let's discuss
- I would maybe make the names of the methods in the profile service more self-explanatory or add docstrings
- What exactly is "identifier", because it's different to the id, no?
- What do we need this commented out code for?

irlaec (changes requested):

octoprint_mrbeam/__init__.py
khaledsherkawi: The list of device series was removed from the plugin implementation because it leads to unneeded overhead every time we add a new series. Now the series comes from the /etc/mrbeam file always.

khaledsherkawi: I will add the Enum for the relevant device series though! Great point

octoprint_mrbeam/__init__.py
khaledsherkawi: This is fine as we can add any relevant arguments to this later. This is similar to other event handlers we are using elsewhere.

khaledsherkawi: I will remove from docstring though

octoprint_mrbeam/__init__.py
khaledsherkawi: On point. I will implement it.

octoprint_mrbeam/__init__.py
khaledsherkawi: I will implement it.

octoprint_mrbeam/__init__.py
khaledsherkawi: I will implement it.

octoprint_mrbeam/constant/profile/laser_cutter/__init__.py
khaledsherkawi: I will implement.

octoprint_mrbeam/constant/profile/laser_cutter/series_2c.py
khaledsherkawi: I will implement.

octoprint_mrbeam/printing/printer.py
khaledsherkawi: Yes, this is an attribute we are overriding from OctoPrint. I will add a comment

octoprint_mrbeam/printing/printer.py
khaledsherkawi: No this is left on purpose for later when we want to remove the restart/reload logic

octoprint_mrbeam/printing/printer.py
khaledsherkawi: I will implement this.

octoprint_mrbeam/service/profile/laser_cutter_profile.py
khaledsherkawi: I will implement this.

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is basically the parent class that will always be implemented to be used and not directly used. So the values here are just for demonstration

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: This is not my implementation but OctoPrint's. We are using their class as a parent class with minor changes to fir our purpose

octoprint_mrbeam/service/profile/profile.py
khaledsherkawi: @irlaec why is this left unresolved?

irlaec: I didn't notice it 😅

irlaec (approved):

